PR #5, opened 2022-10-03
Improve the code health of arnis

callumfrance:
Hey @louis-e,

This seems like a really cool project so here are some things you might want to try out.

* Added a `src/` directory to keep code clean and manageable
    * I have broken down the `arnis.py` file into several smaller files
* Added a auto-code formatter called `black` to style the code consistently (run with `make style` and check with `make style-check`)
* Added `flake8` linter to check that the code is formatted nicely (run with `make lint`)
    * Config is in `.flake8` but you shouldn't need to worry about that too much
* Added a `Makefile` so that you can add your own custom commands to `arnis`
* Added a `.gitignore` so any python-adjacent files that you do not need to worry about will not be stored in GitHub
* Fixed some lint errors

Full disclosure - I have not ran the program so take this pull request with a grain of salt. I haven't actually edited the source code, mostly just moved it into new files as well has ran the linter and auto-formatter on it, but you never know.


louis-e:
Amazing! Thanks a lot for your effort, I really appreciate it.

callumfrance:
No problem

Just reiterating that I did not test this code that you have now merged...

I think I already did make an error - the `/arnis.py` file references the `src/main.py` file by calling a `run()` function. The `run()` function does not exist.
So two options -

- Ignore the `arnis.py` file and run the code straight from `src/main.py`. This is the quick-fix solution.
- Put the code inside of `src/main.py` into functions, most importantly with a `run()` function that becomes the entrypoint and start of the program. This may take a bit of time but you should get better code from it as a result.

louis-e:
No worries, I tested it before merging it. It was nevertheless working. But thanks for pointing that out, I'll fix it!

Issue #16, opened 2022-10-23
Multithread?


HackerLacker:
Generating big cities is kinda of a hell, because its using only one core. It's stuck on "Calculating layers" while recreating Saint-Petersburg.

HackerLacker:
Thanks for answer. I can't tell how much i waited, but i managed to sleep before it get to calculating layers step.

amir16yp:
@louis-e I found a way to optimize this software in a way that generating cities as large as in the picture below is possible. by using memory maps instead of storing in memory, you could generate maps as large as you like. 
![arnis-debug-map](https://github.com/louis-e/arnis/assets/96111605/a2299489-19d3-4d35-a80f-2249057fe953)
I'll send a pull request soon

Issue #8, opened 2022-10-05
Memory allocation error


kosssst:
```
>>> python arnis.py --city "Київ" --state "Київ" --country "Україна" --path "E:/kyiv"
Fetching data...
Chosen server: https://overpass.kumi.systems/api/interpreter
Parsing data...
Traceback (most recent call last):
  File "D:\arnis\arnis.py", line 7, in <module>
    from src.main import run
  File "D:\arnis\src\main.py", line 118, in <module>
    imgarray = processData(rawdata, args)
  File "D:\arnis\src\processData.py", line 56, in processData
    img = np.zeros(
numpy.core._exceptions._ArrayMemoryError: Unable to allocate 275. GiB for an array with shape (1181055, 250396, 1) and data type uint8
```
Hi, when I try to run this script I catch this error. Can anyone help me?

basdvnjksnjksvnsdjknK:
obviously you need to buy more RAM

kostyamed:
275 GB?

kosssst:
`>>> python arnis.py --city "Київ" --state "Київ" --country "Україна" --path "D:/kyiv"
Fetching data...
Chosen server: https://z.overpass-api.de/api/interpreter
Parsing data...
Traceback (most recent call last):
  File "D:\arnis\arnis.py", line 10, in <module>
    run()
  File "D:\arnis\src\main.py", line 123, in run
    imgarray = processData(rawdata, args)
  File "D:\arnis\src\processData.py", line 150, in processData
    img = np.zeros(
numpy.core._exceptions._ArrayMemoryError: Unable to allocate 269. GiB for an array with shape (1181164, 244972, 1) and data type uint8`

@louis-e, I checked commit history and and you pointed out that you improved RAM usage, but problem is still here.

PR #30, opened 2024-05-20
Major optimization and overhall

amir16yp:
This changes the following:
- Use memory maps instead of memory to store specific data, this majorly improves optimization and solves #16 #8, allowing for way larger cities to be rendered
- Make the `--state` flag optional
- Adds a progress bar for `Calculating layers...` job
- Add `--downloader` flag for specifying which downloader (from the choice of `wget`, `curl` and `requests` (default is requests)

### DIFF
--- a/src/getData.py
+++ b/src/getData.py
@@ -1,9 +1,36 @@
 import os
+import subprocess
 import requests
+import json
 from random import choice
 
 
-def getData(city, state, country, debug):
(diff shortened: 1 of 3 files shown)


louis-e:
Hi there, thanks a lot for your contribution!
I will test & merge it asap when I'm back home. :)

PR #31, opened 2024-08-03
Update

daniil2327:
Added bbox, different trees, flowers, removed timeout because forest cannot be generated


daniil2327:
If I'm motivated, I'll write a GUI later.
I don't live in England so there may be mistakes in the words :)

louis-e:
BBOX, amazing!
Will look into this asap and merge it. I've been thinking about switching from the prone to errors city / state / country name input method to bbox for a while already, so this might be the first big step into it. Thanks a lot for your contribution, it is very much appreciated!

louis-e:
A quick update for you; I spent the last two days refactoring a big portion of the code. I will merge your PR and then push my changes which are based on yours.
Thank you very much for your contribution! :)

PR #46, opened 2024-08-24
Upgrade to anvil-new

Aeris1One:
`anvil-parser` is deprecated and only supports the old world format, meaning a 1.16+ client have to do an upgrade of the world.
There have been discussions here about replacing it with `anvil-new` which is a fork of `anvil-parser` for the newer world format.


louis-e:
Looks good to me, thanks for your contribution!

PR #74, opened 2024-10-16
fix chunk file in wrong location

scd31:
WIP as I need to clean up the code and fix the progress bar. It's my bed time, I'll try to get that done tomorrow.

There were two separate problems here.

1. We weren't setting the xPos/zPos chunk data, so the chunks were in the wrong positions.
1. We were only saving modified chunks from the reference MCA file, however we need to save them all (so that we can update all the positions)

Fixes #53 and #38

PR #75, opened 2024-10-17
Make north point north

### LINKED ISSUE #50: [DISCUSSION] Why is North not North?
When rendering an area of London why is everything rendered + 45 degrees?

North should be North. Whereas this script renders North as East.

scd31:
Functionally this works as intended. Just wondering about code cleanup/best practices, so I'm leaving it as a draft for now. Also, this is based off of #74 so that should be merged first.

Fixes #50

### DIFF
--- a/src/osm_parser.rs
+++ b/src/osm_parser.rs
@@ -1,7 +1,7 @@
 use crate::args::Args;
 use colored::Colorize;
-use serde_json::Value;
 use serde::Deserialize;
+use serde_json::Value;
 use std::collections::HashMap;
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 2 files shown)


src/osm_parser.rs
```diff
@@ -93,19 +94,20 @@ pub fn parse_osm_data(
...
-            ((bbox_scaled.2 - bbox_scaled.0) * 10 / 100) as f64, // Scale for width (x) is based on longitude difference
-            ((bbox_scaled.3 - bbox_scaled.1) * 14 / 100) as f64, // Scale for length (z) is based on latitude difference
-        )
-    } else {
-        // Latitude difference is greater than or equal to longitude difference
-        (
-            ((bbox_scaled.3 - bbox_scaled.1) * 10 / 100) as f64, // Scale for width (x) is based on latitude difference
-            ((bbox_scaled.2 - bbox_scaled.0) * 14 / 100) as f64, // Scale for length (z) is based on longitude difference
-        )
-    };
```
scd31, line 108: I don't fully understand what this code is for. If we just use scale_factor_* to determine how big the output should be, why does it depend on whether the latitude delta is bigger/smaller than the longitude delta? And what's the difference of 10 and 14?

src/osm_parser.rs
```diff
@@ -37,8 +37,8 @@ fn lat_lon_to_minecraft_coords(
     let (min_lon, min_lat, max_lon, max_lat) = bbox;
 
     // Calculate the relative position within the bounding box
-    let rel_x: f64 = (lon - min_lon) / (max_lon - min_lon);
-    let rel_z: f64 = (lat - min_lat) / (max_lat - min_lat);
+    let rel_x: f64 = 1.0 - (lat - min_lat) / (max_lat - min_lat);
+    let rel_z: f64 = (lon - min_lon) / (max_lon - min_lon);
```
scd31, line 41: I'd like to swap around the (z, x) a few lines below, and adjust this code accordingly, but it'll depend on what exactly scale_factor_x and scale_factor_z do. With the current version of the code, it seems like we're using scale_factor_z to scale the x coordinate and vice versa

PR #76, opened 2024-10-19
Wide rivers

scd31:
Working on adding rivers. Code works as-is, but it's very slow. Takes about 40 minutes to run on my small city. Have some improvements locally but I haven't pushed them up yet.

This required a bit of refactoring so that we could handle relations. I cleaned up some related code at the same time. And this is also based on my other PRs, so I recommend looking at them first (as some of their code is duplicated in this PR)

Before:

freddy_before

After:

freddy_after

This makes a big difference to how my city looks!

### DIFF
--- a/src/data_processing.rs
+++ b/src/data_processing.rs
@@ -47,75 +47,74 @@ pub fn generate_world(
         if args.debug {
             process_pb.set_message(format!(
                 "(Element ID: {} / Type: {})",
-                element.id, element.r#type
+                element.id(),
+                element.kind()
             ));
         } else {
             process_pb.set_message("");
         }
 
-        match element.r#type.as_str() {
-            "way" => {
-                if element.tags.contains_key("building")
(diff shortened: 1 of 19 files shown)


louis-e:
I love your inversed floodfill approach.
Awesome work, thanks for your time and effort! :)

scd31:
@louis-e Sorry, I was out for a wedding late last night and only just got up. I think we should to revert this PR and #75 as they were both draft PRs. This one I'm working on optimizing more (and tbh, it should probably be tested a bit more as I refactored a bunch of code) and the other one I left some questions on in the PR.

scd31:
Actually, we'd probably okay to not revert (I can put up new PRs to finish things) but #75 has a few questions I was wondering about

scd31:
Sorry for all the spam, but my open PRs should fix all the WIP stuff! Can probably ignore my above comments

Issue #63, opened 2024-09-21
[BUG] The new rust version again falls for the same out of memory errors


amir16yp:
**Describe the bug**
Runs out of memory when computing large maps. I think its the same as #8 except the problem is in rust. The solution would be to use memory-map files to store data instead of memory like i did in the python version in PR #30  
**Used bbox parameter**
34.549942,31.490310,34.651909,31.570931

**Screenshots**
![image](https://github.com/user-attachments/assets/7c60da52-085d-4357-bfce-d17c6315d071)

**Additional context**
Issue does not occur in the python version

louis-e:
Hi there Amir, thanks for the notice! Besides a few minor bugs, this is the main thing I'm working on right now. I'm currently investigating what implementation approach is the best for this. Especially for finding a good balance between memory usage and performance. It's not quite easy to implement this feature since the element placement order does not follow a pattern which can be mapped to certain coordinate wise incremeting locations. But I'm certain, I can show off a first solution pretty soon! Already got something running right now.
I will use this Issue for keeping everyone up to date on this matter!

louis-e:
Sorry for the late follow up, but this should've been solved by https://github.com/louis-e/arnis/pull/78 now! Just double checked it with your bbox parameter.

PR #78, opened 2024-10-20
Memory improvements

### LINKED ISSUE #72: I can't create a world. Not enough memory.
After porting to Rust, I can't create a world because creating it takes too much RAM. I get the error “memory allocation of 123920 bytes failed”. And I have 16GB of RAM

scd31:
There is probably more that could be done here, but I'd say this fixes #72 and #63. My city took 40GB of RAM originally, and then 20GB after #74. It now only takes 2-3GB of RAM for the same city.

The main improvement comes from reducing the size of the `Block` struct. It's now only 1 byte long. This has also improved performance a bit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ clap = { version = "4.1", features = ["derive"] }
 colored = "2.1.0"
 fastanvil = "0.31.0"
 fastnbt = "2.5.0"
+fnv = "1.0.7"
 geo = "0.28.0"
 indicatif = "0.17.8"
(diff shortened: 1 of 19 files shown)


louis-e:
Amazing, I can't keep up with my changes since you're so fast haha!
Huge thanks for your contributions. Sorry for being a bit late on merging them, I'm trying to find some more time for Arnis again. I'm really motivated on pushing the development of this project forward to a new release. :)

scd31:
No worries at all!

PR #91, opened 2024-12-28
Added settings menu including scale option and custom bbox input

louis-e:
image


faithsapling:
What does the Custom Bounding Box do?

PR #98, opened 2024-12-29
Added support for generating new world

louis-e:
image

image


Troxter222:
I have some problem:

Error: Minecraft directory not found.

How can i give directory to my minecraft

PR #105, opened 2024-12-31
Fixed longitude wrapping bug

louis-e:


louis-e:
Needs further testing, not working as expected

PR #111, opened 2025-01-01
Add block letters for street names

IAmStoxe:
Add functionality to generate and display block letters for street names.

* **src/element_processing/highways.rs**
  - Add a new function `generate_block_letters` to generate block letters for street names.
  - Update `generate_highways` function to call `generate_block_letters` when generating street names.

* **gui-src/js/main.js**
  - Add a new function `displayBlockLetters` to handle the display of block letters for street names on the map.
  - Update `registerMessageEvent` function to call `displayBlockLetters` when receiving block letter data.

### DIFF
--- a/src/element_processing/highways.rs
+++ b/src/element_processing/highways.rs
@@ -264,6 +264,11 @@ pub fn generate_highways(
                 }
                 previous_node = Some((node.x, node.z));
             }
+
+            // Generate block letters for street names
+            if let Some(name) = element.tags().get("name") {
+                generate_block_letters(editor, name, way, ground_level);
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/element_processing/highways.rs
```diff
@@ -297,3 +302,252 @@ pub fn generate_siding(editor: &mut WorldEditor, element: &ProcessedWay, ground_
         previous_node = Some((x, z));
     }
 }
+
+/// Generates block letters for street names
+fn generate_block_letters(
```
benjamin051000, line 307: Could this be done with a lookup-table or something? What about street names that don't use the English alphabet? We may need a more efficient solution.

PR #136, opened 2025-01-03
Enable generation without having Minecraft installed

louis-e:
When selecting 'Select existing world', a custom directory can be selected without the need to have Minecraft installed.


Dimaa1113:
add custom Minecraft directory for world generation, i am use MultiMC, and add search places on the map

PR #170, opened 2025-01-04
POC Street names as signs

louis-e:
Needs some more testing, generates faulty artifacts
image

image

### DIFF
--- a/src/element_processing/highways.rs
+++ b/src/element_processing/highways.rs
@@ -102,6 +102,41 @@ pub fn generate_highways(
                 editor.set_block(surface_block, x, ground_level, z, None, None);
             }
         } else {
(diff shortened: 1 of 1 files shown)


wielandb:
You mentioned in #214 that you are not happy with how the street name signs look. 
I am not really a good builder myself, but as I want to try to add guideposts eventually, I have experimented a bit.
What do you think about a design that is something like this:





It would also have the added benefit of not restricting the signs to 4 directions like you would normally have when attaching signs to a pole.

louis-e:
Sorry for the delay, I kind of forgot about this PR! :D
No updates yet, but I can continue my work of supporting signs.

suprohub:
Bruh, still waiting hehe

PR #230, opened 2025-01-07
License Change from GPL-3.0 to Apache 2.0

louis-e:
**Summary:**
This pull request proposes changing the project's license from GPL-3.0 to Apache License 2.0. The primary motivation for this change is to address ongoing issues with unauthorized websites and individuals misusing the project's name and branding without proper attribution or permission. The Apache 2.0 license offers additional protections, including a trademark clause that prevents misuse of the project name in a way that could mislead users.

**Reason for the Change:**
We've recently encountered several instances of unauthorized websites falsely presenting themselves as official sources of Arnis. While the GPL-3.0 license ensures the project's openness, it lacks provisions to prevent brand misuse. The Apache 2.0 license requires users to give proper attribution and explicitly prohibits using the project's name or branding without permission.

The project will remain open source and free for everyone to use, modify, and contribute to, as it has always been.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,674 +1,201 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 3, 29 June 2007
-
- Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
- Everyone is permitted to copy and distribute verbatim copies
- of this license document, but changing it is not allowed.
-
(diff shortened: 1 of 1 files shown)


TheComputerGuy96:
The problem with Apache 2.0 is that it lacks copyleft (which allows people to make closed-source clones of this program)

PR #240, opened 2025-01-07
Add generic linux binary to releases workflow

amir16yp:
This will allow easier packaging in of the binary version in various linux distros (it compiles it using an ubuntu runner, but i have found that the binary works on arch and probably any glibc distro with xorg)


louis-e:
Looks good to me, thank you so much! :)

PR #258, opened 2025-01-10
Ignore barrier=kerb

RedAuburn:
before:
Screenshot From 2025-01-10 22-50-40

after:
Screenshot From 2025-01-10 22-47-07


louis-e:
Looks good to me, thanks a lot! :)
Just fixed the recent Clippy lint check CI fails, so I'm just waiting for this one to run sucessfully and then I'll merge it!

PR #259, opened 2025-01-10
Use Minecraft Flatpak directory if it exists

RedAuburn:
Makes selecting saves or generating a new one work when minecraft is installed as a flatpak on Linux

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -175,7 +175,14 @@ fn gui_select_world(generate_new: bool) -> Result<String, i32> {
                 .join("saves")
         })
     } else if cfg!(target_os = "linux") {
-        dirs::home_dir().map(|home: PathBuf| home.join(".minecraft").join("saves"))
+        dirs::home_dir().map(|home| {
+            let flatpak_path = home.join(".var/app/com.mojang.Minecraft/.minecraft/saves");
+            if flatpak_path.exists() {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -175,7 +175,22 @@ fn gui_select_world(generate_new: bool) -> Result<String, i32> {
                 .join("saves")
         })
     } else if cfg!(target_os = "linux") {
-        dirs::home_dir().map(|home: PathBuf| home.join(".minecraft").join("saves"))
+        let flatpak_path = |home: &PathBuf| {
+            home.join(".var")
+                .join("app")
+                .join("com.mojang.Minecraft")
+                .join(".minecraft")
+                .join("saves")
```
benjamin051000: Is there any way to combine these into one string?

benjamin051000:
Nice addition. See comments

src/main.rs
hb0nd: yup, it can just be passed as one - i thought there might be some reason it was previously done with join() though so i just kept using it

RedAuburn:
simplified

louis-e:
Looks good to me, thanks everyone!
Just fixed the recent Clippy lint check CI fails, so I'm just waiting for this one to run sucessfully and then I'll merge it!

Issue #159, opened 2025-01-04
[FEATURE] Make the GUI an optional cargo feature


benjamin051000:
There are two ways to interact with Arnis: The GUI and the CLI.

The GUI is nice, and is the default way to interact with the project. However, I honestly prefer the CLI, and I'm sure others do, too. #linuxuser 

It would be nice if I could strip out the GUI components for faster compile times/reduced dependencies. I believe the Rust way of doing this is by making CLI and GUI "features" in `Cargo.toml`?

I still think the GUI should be the default, but you could perhaps make it possible to disable it and get only the CLI.

That way I don't have to compile like 800 deps and can have a slim version of the app. :D 

Also, there could be a way to strip out the CLI and just have the GUI, although I'm not sure how much compile/space savings that would offer since the CLI is pretty simple.

louis-e:
The CLI mode was originally even the standard interface, it's still in the code! Just remove this line which suppresses it: https://github.com/louis-e/arnis/blob/2f2d1c79bb1f7c2e8cb67d96dcdf6f4610cc8591/src/main.rs#L1

And run it as follows: arnis --path="C:/YOUR_PATH/.minecraft/saves/worldname" --bbox="min_lng,min_lat,max_lng,max_lat"

We're currently discussing in https://github.com/louis-e/arnis/issues/99 if we really want to keep the command line interface. And if so, how we should make it available on Github. I agree with you, it'd be cool to have a stripped down version with just the CLI - especially for making the compilation a bit easier on Linux. However I'm not sure yet how to do it while keeping an easy maintainability

benjamin051000:
I mean, it's definitely not a deal-breaker. Just a nice-to-have. If only 1% of users want a command line, it may not be worth the effort. 🤷🏼‍♂️ Feel free to focus on other efforts in the mean-time

austenadler:
I made a PR here: https://github.com/louis-e/arnis/pull/260/files?w=1. You can just run `cargo build --no-default-features` and tauri is disabled.

louis-e:
Closing now since this is integrated! :)

PR #260, opened 2025-01-11
Allow disabling gui at compile time

austenadler:
Allows users to disable the GUI at compile time by using `--no-default-features`


README.md
```diff
@@ -93,9 +93,9 @@ Feel free to choose an item from the To-Do or Known Bugs list, or bring your own
...
+This project is open source and welcomes contributions from everyone! Whether you're interested in fixing bugs, improving performance, adding new features, or enhancing documentation, your input is valuable. Simply fork the repository, make your changes, and submit a pull request. We encourage discussions and suggestions to ensure the project remains modular, optimized, and easy to use for the community. You can use the parameter --debug to get a more detailed output of the processed values, which can be helpful for debugging and development. Contributions of all levels are appreciated, and your efforts help improve this tool for everyone. You can add `--no-default-features` to disable the GUI (enabled by default).
```
benjamin051000: This shouldn't be in the "how to contribute" section. It's more of a "how to use arnis" thing.

benjamin051000 (commented):
I did not do code review, but nice work. I can test this in two days when I'm home. 

We definitely should discuss how this might work with releases... having two executables, one with gui, one without, is kinda weird. Maybe this shouldn't be a compile feature... I don't know. Maybe the gui should live in its own repo. 🤷‍♂️ things to discuss I guess

benjamin051000:
Oh, this also needs CI testing. We need to ensure it compiles with and without gui. Can you add that to the github workflow?

benjamin051000:
Resolves #159

amir16yp:
if this is merged i can submit a PR that will add CLI-only version to releases

louis-e:
Sounds very good to me, I'll look into this asap! :)

louis-e:
A few thoughts from my side, I like this implementation and I think it makes sense to do it in this way. Sadly there seems to be no way to make this line, which suppresses the command line, optional for a single release:
https://github.com/louis-e/arnis/blob/a92c5bb01f28662987d20fce0bec3042a67924bc/src/main.rs#L1

That's why I'm okay with a seperate CLI only release asset, as long as it's clearly distinguishable via the asset name (e.g. "arnis-cli.exe").
Also I somehow can't get this to run in CLI only mode, can you confirm that this is the correct way to execute it?
image

austenadler:
> Sadly there seems to be no way to make this line, which suppresses the command line, optional for a single release

> Also I somehow can't get this to run in CLI only mode, can you confirm that this is the correct way to execute it?

Hmm I am not on Windows and I didn't try it there. I will check it out today and see why it's failing for windows builds. It might be the windows_subsystem line that's preventing an actual error from being printed?

README.md
austenadler: Good call; I removed it

austenadler:
There are a few points here. 

1: Your coordinates aren't generating any data, which is why it exited unsuccessfully. The output should have been:

```
[1/5] Fetching data...
Error! No data available in this region.
error: process didn't exit successfully: `target\release\arnis.exe --path=C:\Users\user\Downloads\superflat --bbox=48.139631,11.564054,48.145014,11.574440` (exit code: 1)
```

It works if you use a different set of coordinates (for example: `cargo r -r -- "--path=C:\Users\user\Downloads\superflat" --bbox=-80.849226,34.117578,-80.841286,34.123032`

2: Windows does not send anything to stdout/stderr if you use `windows_subsystem = "windows"` (by default). So you were not seeing that error message in Windows, but it would show up for linux. According to [this](https://github.com/rust-lang/rust/issues/67159#issuecomment-987882771) and [this](https://old.reddit.com/r/learnrust/comments/jaqfcx/windows_print_to_hidden_console_window/hyse729/?context=1), you can reattach to the parent console at runtime.

If you double click the .exe, you get the GUI. If you run `.\arnis.exe`, you get the GUI. If you specify `--path=... --bbox=...`, you get stdout. It's the best of both worlds. 

3: According to the [FreeConsole](https://learn.microsoft.com/en-us/windows/console/freeconsole) and [AttachConsole](https://learn.microsoft.com/en-us/windows/console/attachconsole) documentation, you can safely free if not already attached, and I don't see any warnings about using these functions, so they should be fine to use even though they have to be in an `unsafe` block.

I pushed the fix to this. Everything should work now; please review again.

louis-e:
Thanks a lot for your great work and sorry for catching up with it a bit late!
I've now reviewed and tested everything, this works perfectly! Thank you for your effort :)

PR #271, opened 2025-01-12
Faster saving

louis-e:


louis-e:
Replaced by https://github.com/louis-e/arnis/pull/272

PR #275, opened 2025-01-12
fix input text colour

RedAuburn:
before/after
image1


TheComputerGuy96:
#290 contains a different approach for this (which enables the dark input box background color too)

RedAuburn:
> #290 contains a different approach for this (which enables the dark input box background color too)

thanks! i didn't look into fixing it too hard - yours looks better

PR #286, opened 2025-01-13
Support barrier / fence types

RedAuburn:
Adds support for some barrier and fence types

before/after 
image1


RedAuburn:
added more `fence_type`s

louis-e:
Looks good, thanks a lot! :)

PR #289, opened 2025-01-14
Translation support rework

TheComputerGuy96:
This also adds a Lithuanian translation (it may not be the best though)


louis-e:
Sorry for the late response, but this is really great - thank you very much for your effort! :)

PR #290, opened 2025-01-14
Miscellaneous fixes

### LINKED ISSUE #244: [DISCUSSION] Failed to create GBM buffer of size 1000x650: Invalid argument
Just getting error from Terminal when I try to run this binary file from releases workflow with given chmod +x rights. Surfing the such issues, found another in https://github.com/tauri-apps/wry/issues/1366#issuecomment-2369391532. Probably this is related to Arch Linux distro I use.

Seems that appeared not only for me, this also happen in #106 when someone's trying to compile for Linux.

TheComputerGuy96:
Fixes #244
Supersedes #275

### DIFF
--- a/gui-src/js/main.js
+++ b/gui-src/js/main.js
@@ -452,7 +452,7 @@ function handleWorldSelectionError(errorCode) {
   document.getElementById('selected-world').textContent = errorMessage;
   document.getElementById('selected-world').style.color = "#fa7878";
   worldPath = "";
-  console.error(error);
+  console.error(errorCode);
 }
 
 let generationButtonEnabled = true;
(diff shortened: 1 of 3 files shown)


louis-e:
Thanks a lot for your research and fixes for these bugs! :)

gui-src/js/main.js
```diff
@@ -1,6 +1,11 @@
 import { licenseText } from './license.js';
 
-const { invoke } = window.__TAURI__.core;
+if (window.__TAURI__) {
```
szilard-dobai, line 3: Sorry, but this change is causing an issue when trying to load world due to `invoke` now being scoped only inside the `if` block. You'll need to create a `let` variable above (in the global scope), for example, and assign it the value of this `invoke`.

image
(screenshot after changing the code a tiny bit to display the `errorCode`)

@TheComputerGuy96 @louis-e

TheComputerGuy96: #296 should probably fix this

PR #296, opened 2025-01-16
Use global scope for invoke definition

TheComputerGuy96:
Fixes: dce5f47 ("Make the GUI partially work in a web browser")


TheComputerGuy96:
Fixed in a different way by fc105b84e0b9afb0c0e5b7d1c103b93da6872cf4

PR #306, opened 2025-01-19
Add "razed" to values of railway=* not rendered

wielandb:
In my region, there are a lot of historic but not anymore existing railways tagged with railway=razed. Theese should not be rendered as well.


wielandb:
Hmmmm.... the checks fail because of a formatting error. Curiously, the formatting error that the check found was not what I committed, as is visible in the changed files section.

RedAuburn:
i agree the lint check reports are confusing, but it's in a diff format (`+`/`-`).

Here it is formatted in github markdown:

```diff
-        if ["proposed", "abandoned", "subway", "construction", "razed"].contains(&railway_type.as_str()) {
+        if ["proposed", "abandoned", "subway", "construction", "razed"]
+            .contains(&railway_type.as_str())
+        {
             return;
         }
```

other than that, the change looks good :)

wielandb:
Oh, so the lines marked with `+` are how it _should_ be! Do I understand correct that I should change it to the way the check suggested?

RedAuburn:
> Oh, so the lines marked with `+` are how it _should_ be! Do I understand correct that I should change it to the way the check suggested?

yup that will make the check pass :)

wielandb:
Thanks for the clarification! Changed it. ✔

wielandb:
While exploring, I noticed that the value "turntable" should probably also not be rendered. Added it.

louis-e:
Thanks a lot for your contribution! :)

doruk3434:
nasıl  haritayı açıp seçecez

PR #313, opened 2025-01-21
Add CLI build test to CI

benjamin051000:
Now that we can build a separate CLI version of the app, we should verify it compiles in the CI build.


benjamin051000:
This definitely takes a while for CI to complete. Would it be better if this was in its own file, so they could run concurrently? Or maybe there's a way to tell GH Actions that the steps can run in parallel?

benjamin051000:
Whoops. I should have made this a feature branch. I'll reopen

PR #320, opened 2025-01-22
Add CLI build test to CI

benjamin051000:
Since the gui is now an optional feature, we should also test that the CLI builds on its own.


louis-e:
Hi Benjamin, thanks a lot for all your contributions! Sorry for being a bit late with responding to all of them :D
Do you think this will be necessary? I think the chance of a CLI only build failing when the GUI build passes will be pretty low. Just asking because I recently received an email by Github telling me that I already used 75% of my included Actions runtime and this PR will increase it even more :D

PR #321, opened 2025-01-22
Remove gui-only deps when building cli

benjamin051000:
It looks like these are only GUI dependencies. I still need to test that it compiles, but these give unused warnings when compiling with the gui disabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,15 +13,15 @@ lto = "thin"
 
 [features]
 default = ["gui"]
-gui = ["dep:tauri", "dep:tauri-plugin-log", "dep:tauri-plugin-shell"]
+gui = ["tauri", "tauri-plugin-log", "tauri-plugin-shell", "tokio", "rfd", "dirs", "tauri-build"]
 
 [build-dependencies]
-tauri-build = "2"
+tauri-build = {version = "2", optional = true}
 
 [dependencies]
 clap = { version = "4.1", features = ["derive"] }
 colored = "2.1.0"
-dirs = "5.0.1"
+dirs = {version = "5.0.1", optional = true }
 fastanvil = "0.31.0"
 fastnbt = "2.5.0"
 flate2 = "1.0"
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -28,18 +28,23 @@ mod world_editor;
...
+#[cfg(feature = "gui")]
 use log::{error, LevelFilter};
+#[cfg(feature = "gui")]
 use rfd::FileDialog;
 use std::{
     env,
-    fs::{self, File},
-    io::{Read, Write},
+    fs,
+    io::Write,
```
benjamin051000: I may need to include `io::Read` for GUI functionality.

src/main.rs
```diff
@@ -125,8 +130,8 @@ fn main() {
 
         // Write the parsed OSM data to a file for inspection
         if args.debug {
-            let mut output_file: File =
-                File::create("parsed_osm_data.txt").expect("Failed to create output file");
+            let mut output_file: fs::File =
+                fs::File::create("parsed_osm_data.txt").expect("Failed to create output file");
```
benjamin051000, line 131: I found this simpler than `use std::fs::File` when playing with the `use` statements above. It can probably be changed if need be.

Cargo.toml
```diff
@@ -13,7 +13,7 @@ lto = "thin"
 
 [features]
 default = ["gui"]
-gui = ["dep:tauri", "dep:tauri-plugin-log", "dep:tauri-plugin-shell"]
+gui = ["dep:tauri", "dep:tauri-plugin-log", "dep:tauri-plugin-shell", "dep:tokio"]
```
benjamin051000: I think it's only used with functions involving Tauri.

benjamin051000 (commented):
Let me know what you think. This removes unused warnings during `cargo build --no-default-features`.

benjamin051000:
Right now main has 490 dependencies, this branch reduces it to 241 😎 My machine compiles it in 25s

louis-e:
Hey Benjamin, I'll work through a few if not all of your pull requests tonight. Thanks a lot for your amazing effort in the past few weeks and sorry for catching up with it a bit late! :)

benjamin051000:
No worries @louis-e ! I'm glad you're back :D I took a break as well

PR #327, opened 2025-01-23
Change what trees are generated based on tag content

wielandb:
This PR makes trees (`natural=tree`) and trees in woods (`natural=wood`) and forests (`landuse=forest`) generate depending on infos from tags.

If a `species` or `species:wikidata` is given and it is one of the trees we can generate, exactly that one is generated. (`natural=tree` only)
If `leaf_type=broadleafed/needleleafed` is given, only trees/a tree that has leafs or needles respectively is generated.


Oleg4260:
No activity for 3 months... did you forget about your PR?

PR #334, opened 2025-01-24
Make leaves persistent to stop hedges from decaying

### LINKED ISSUE #329: [BUG] Leaves from hedges are decaying
Leaves that are part of a hedge decay over time. This happens if the leaves are considered to be naturally generated, if they are player-placed or a log is near them, this wont happen.

I drafted a change to always place an oak log underneath every hedge segment, but this may not be the best solution. If the block state `persistent` could be set to `true`, theese leaves wont decay.

Is the world editor currently capable of setting block states? (If yes, I would be happy to create a PR 🙂)

wielandb:
This PR makes the "OAK_LEAVES" Block always have `persistent` set to `true`, so that they don't decay.
Fixes #329

### DIFF
--- a/src/block_definitions.rs
+++ b/src/block_definitions.rs
@@ -150,6 +150,11 @@ impl Block {
 
     pub fn properties(&self) -> Option<Value> {
         match self.id {
+            49 => Some(Value::Compound({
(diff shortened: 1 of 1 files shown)


louis-e:
Great, thanks a lot! :)

PR #335, opened 2025-01-25
Refactor tree generation

benjamin051000:
I'm learning the code base by refactoring small files at a time. Tree generation was not so intimidating, so I started here. 😄 

I intend on reducing repetition in this file, making it easier to comprehend and lessening the chance of making a mistake.

What do you think so far? I want to go even further by reducing each of the match arms, since they all kinda do the same thing. I'll get to it tomorrow.

I would probably recommend reviewing this commit-by-commit, since the overall diff will be large.

### DIFF
--- a/src/element_processing/tree.rs
+++ b/src/element_processing/tree.rs
@@ -2,7 +2,7 @@ use crate::block_definitions::*;
 use crate::world_editor::WorldEditor;
 
 /// A circular pattern around a central point.
-const ROUND1_COORDS: [(i32, i32, i32); 8] = [
+const ROUND1_PATTERN: [(i32, i32, i32); 8] = [
     (-2, 0, 0),
     (2, 0, 0),
     (0, 0, -2),
@@ -14,7 +14,7 @@ const ROUND1_COORDS: [(i32, i32, i32); 8] = [
 ];
 
 /// A wider circular pattern.
-const ROUND2_COORDS: [(i32, i32, i32); 12] = [
+const ROUND2_PATTERN: [(i32, i32, i32); 12] = [
     (3, 0, 0),
     (2, 0, -1),
     (2, 0, 1),
@@ -30,7 +30,7 @@ const ROUND2_COORDS: [(i32, i32, i32); 12] = [
 ];
 
 /// A more scattered circular pattern.
-const ROUND3_COORDS: [(i32, i32, i32); 12] = [
+const ROUND3_PATTERN: [(i32, i32, i32); 12] = [
     (3, 0, -1),
     (3, 0, 1),
     (2, 0, -2),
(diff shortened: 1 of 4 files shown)


benjamin051000:
Currently, even though the line count went up by about 40 lines (not sure how...), the character count went down by about 2000 chars!

wielandb:
Could you take a look at #327 ? Does it interfere with your refactorings? Should I change anything?

benjamin051000:
@wielandb looks fine! You didn't edit the tree.rs file. The only thing you'll need to change is a slight modification to the `create_tree` signature (x y z are now a Tuple). Otherwise looks fine and should have no problem merging.

src/element_processing/tree.rs
```diff
@@ -57,108 +65,143 @@ pub fn create_tree(editor: &mut WorldEditor, x: i32, y: i32, z: i32, typetree: u
...
+            let birch_leaves_fill = [
+                ((-1, 3, 0), (-1, 10, 0)),
+                ((0, 3, -1), (0, 10, -1)),
+                ((1, 3, 0), (1, 10, 0)),
+                ((0, 3, -1), (0, 10, -1)),
+                ((0, 3, 1), (0, 10, 1)),
+            ];
+            for ((i1, j1, k1), (i2, j2, k2)) in birch_leaves_fill {
+                editor.fill_blocks(BIRCH_LEAVES, x + i1, y + j1, z + k1, x + i2, y + j2, z + k2, None, None);
+            }
```
benjamin051000: Why does the spruce tree get Birch leaves? Shouldn't it get spruce leaves? Is this intentional?

benjamin051000:
@wielandb Actuallllyyyyyy I might change it more XD But it still shouldn't be a problem because you only call `create_tree()` in a few spots in your PR so at most you'll only have to change a few lines. 99% of this PR is in `tree.rs` which you didn't touch.

Oleg4260:
It gives compilation error because of bad formatting, execute `cargo fmt` and do another commit

benjamin051000:
@wielandb the API has changed. See `Tree::create()` for details. It's actually simpler now, because the RNG is handled internally. Should be very easy to change wherever you need in your branch (assuming this gets merged).

wielandb:
> @wielandb the API has changed. See `Tree::create()` for details. It's actually simpler now, because the RNG is handled internally. Should be very easy to change wherever you need in your branch (assuming this gets merged).

Ah, okay! I will mark my PR as a draft until this is (hopefully) merged.

louis-e:
> Ah, okay! I will mark my PR as a draft until this is (hopefully) merged.

Finally merged. You'll need to resolve a few merge conflicts, then I'll immediately merge yours. Sorry for the delay and thanks for the great work! :)

PR #337, opened 2025-01-25
Added ores generation in quarries + other changes

### LINKED ISSUE #330: [FEATURE] Implement quarries with stone and ore
**Is your feature request related to a problem? Please describe.**
There has to be a new type of terrain implemented. I really like the new elevation feature, I waited for it for a long time, and I wanted to see how quarries look like, but they turned out to look like huge pits with grass.

**Describe the solution you'd like**
There is an object type "quarry" in the OSM, also there's tag "resource", so this data can be used to generate stone blocks instead of grass and also small bits of ore can be added, it would be very cool

**Additional context**
This is the bbox I use: 47.824796, 33.286858, 47.851759, 33.328571

Image
Image

Oleg4260:
In my previous pull I only added landuse for quarries and landfills, so there would be stone in quarries, but I did not add ores as said in #330, now I added ores as well

Fixes #330


benjamin051000 (approved):

Oleg4260:
I just added another feature that can be enabled in the settings. It fills the underground space with stone, I think it can be useful for many users, but it also should be disabled by default because it increases generation time significantly. Only applies when terrain is enabled.
image

src/data_processing.rs
```diff
@@ -170,6 +170,13 @@ pub fn generate_world(
                 editor.set_block(groundlayer_block, x, max_y, z, None, None);
                 editor.set_block(DIRT, x, max_y - 1, z, None, None);
                 editor.set_block(DIRT, x, max_y - 2, z, None, None);
+                // Fill underground with stone
+                if args.fillground {
+                    for y in MIN_Y + 1..max_y - 2 {
+                        editor.set_block(STONE, x, y, z, None, None);
+                    }
```
benjamin051000: Use `editor.fill_blocks()` instead.

src/element_processing/landuse.rs
```diff
@@ -349,6 +350,28 @@ pub fn generate_landuse(
...
+                    };
+                    let random_choice: i32 = rng.gen_range(0..100 + ground_level); // with more depth there's more resources
+                    if random_choice < 5 {
+                        editor.set_block(ore_block, x, ground_level, z, Some(&[STONE]), None);
+                    }
+                    // Fill everything with stone so dirt won't be there
+                    if args.fillground {
+                        for y in MIN_Y + 1..ground_level {
+                            editor.set_block(STONE, x, y, z, None, None);
+                        }
```
benjamin051000: `editor.fill_blocks()`

benjamin051000 (approved):
I didn't review the JS but it looks fine. The Rust is fine as well but consider using `fill_blocks()` to make it a little simpler

benjamin051000 (changes requested):
Please squash the last commit into the previous one. You can do this with an interactive rebase (`git rebase -i HEAD~2`). Do either a squash or a fixup on the last one so that the previous 2 commits get combined

Oleg4260:
I am not sure if I did it right, let's just hope I didn't break anything...

Oleg4260:
@benjamin051000 thank you for your advice, this is really better and also ground generation feels faster now.

Oleg4260:
Looks like the checking proccess broke for some reason and I can see it in other pull requests as well. Should I do something with it?

benjamin051000:
See #378

Oleg4260:
I'm sorry if this commit wasn't necessary, it's my first time dealing with merge conflicts, should I do something or can the maintainer fix it?

Oleg4260:
Ok, i think i figured it out. But i feel awkward making a lot of commits

benjamin051000:
You can squash the extra commits together using `git rebase --interactive` if you want to. 

But there are loads of commits in main that are typos, merges, not actual changes. So it's probably fine, don't worry about it unless @louis-e says otherwise lol 👍

src/args.rs
```diff
@@ -44,7 +44,9 @@ pub struct Args {
     /// Enable terrain (optional)
     #[arg(long, default_value_t = false, action = clap::ArgAction::SetFalse)]
     pub terrain: bool,
-
+    /// Enable filling ground (optional)
+    #[arg(long, default_value_t = false, action = clap::ArgAction::SetFalse)]
```
benjamin051000, line 48: Is setfalse what we want here? I don't think this will ever be true.

benjamin051000 (changes requested):
Unsure about that argument. See #374 where I had to change that, and added unit tests. If #374 gets merged you should add a unit test too to make sure the argument gets parsed correctly

louis-e:
Hi there, I have to apologize for the long delay - but looks good to me, thanks a lot for your contribution! :)

PR #340, opened 2025-01-26
Change functions returning vecs to global arrays.

benjamin051000:
I'm not sure why these were initally functions. I think they are better off as global arrays, since the function would initialize a new vec every time its called, whereas this only initializes it once.

TODO
- [x] Test: I tested it with one small section of a city and it appears to work. I wouldn't by any means call this thoroughly-tested.

### DIFF
--- a/src/block_definitions.rs
+++ b/src/block_definitions.rs
@@ -427,95 +427,89 @@ pub const PACKED_ICE: Block = Block::new(134);
 pub const MUD: Block = Block::new(135);
 
 // Variations for building corners
-pub fn building_corner_variations() -> Vec<Block> {
-    vec![
-        STONE_BRICKS,
-        COBBLESTONE,
-        BRICK,
-        MOSSY_COBBLESTONE,
-        SANDSTONE,
-        RED_NETHER_BRICKS,
-        BLACKSTONE,
-        SMOOTH_QUARTZ,
-        CHISELED_STONE_BRICKS,
-        POLISHED_BASALT,
-        CUT_SANDSTONE,
-        POLISHED_BLACKSTONE_BRICKS,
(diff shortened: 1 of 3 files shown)


src/block_definitions.rs
```diff
@@ -413,95 +413,89 @@ pub const IRON_ORE: Block = Block::new(127);
...
-        GRANITE,
-        DIORITE,
-        CRACKED_STONE_BRICKS,
-        PRISMARINE,
-        BLUE_TERRACOTTA,
-        NETHER_BRICK,
-        QUARTZ_BRICKS,
-    ]
-}
+pub static BUILDING_CORNER_VARIATIONS: [Block; 20] = [
```
benjamin051000, line 430: It may be worth considering making these `const` instead of `static`, but I'm not sure. In the future, someone can see if it makes a performance difference.

benjamin051000:
This is mostly overshadowed by #351. We can keep it open or merge it, but when #351 is ready for review that one will modify what's going on here. lol

louis-e:
> This is mostly overshadowed by #351. We can keep it open or merge it, but when #351 is ready for review that one will modify what's going on here. lol

Both approaches look good to me and seem to be a better implementation compared to the current state. Would you prefer https://github.com/louis-e/arnis/pull/351 over this one? There isn't any time pressure from my side since the next release still needs a few more feature merges, so we could work on finishing #351. Having a better solution than the constant required unwrap calls would be great, like you already mentioned. Really cool work and documentation from your side btw 😎

benjamin051000:
@louis-e Once CI passes, let's merge this. I don't know when/if I'll finish #351 so may as well merge this work I suppose

PR #351, opened 2025-01-29
Blocks toml

benjamin051000:
The current `block-defs.rs` is not a great solution. Instead, put all block info in a toml file, and pull it in (ideally, at compile time). Then, simply query the one you need either by its ID or name.


src/block_definitions.rs
```diff
@@ -7,501 +7,56 @@ use std::collections::HashMap;
...
+pub struct Block {
+    name: String,
+    id: u8, // TODO what type should this be?
+    properties: Option<HashMap<String, PropertyType>>,
+    building_corner: bool, // TODO may need Option
+
+    /// https://wiki.openstreetmap.org/wiki/Key:building:colour
+    wall_color: Option<[u8; 3]>,
+    floor_color: Option<[u8; 3]>,
+}
```
benjamin051000, line 27: The new Block struct contains all the info from the TOML file. Note that `wall_color` and `floor_color` are `Option`s (and maybe `building_corner`).

src/block_definitions.rs
```diff
@@ -7,501 +7,56 @@ use std::collections::HashMap;
...
 
-    pub fn id(&self) -> u8 {
-        self.id
-    }
+#[derive(Deserialize, Debug)]
+struct Blocks {
+    blocks: Vec<Block>
+}
```
benjamin051000: Apparently this wrapper struct is something you just have to do when deserializing an [array of tables](https://toml.io/en/v1.0.0#array-of-tables). 🤷🏼‍♂️ I'm fine with it for now.

src/block_definitions.rs
```diff
@@ -7,501 +7,56 @@ use std::collections::HashMap;
...
-
-            125 => Some(Value::Compound({
-                let mut map = HashMap::new();
-                map.insert("shape".to_string(), Value::String("south_west".to_string()));
-                map
-            })),
-            _ => None,
-        }
+    pub fn by_id(&self, id: u8) -> Option<Block> {
+        self.blocks.iter().find(|e| e.id == id)
```
benjamin051000: So, how do we reference a block? This is my first attempt. As you'll see, it's a bit verbose with the Option. I'm open to suggestions.


The `.find` should be fast enough since there are only hundreds of blocks.

src/block_definitions.rs
```diff
@@ -7,501 +7,56 @@ use std::collections::HashMap;
...
-        CRACKED_STONE_BRICKS,
-        PRISMARINE,
-        BLUE_TERRACOTTA,
-        NETHER_BRICK,
-        QUARTZ_BRICKS,
-    ]
+    block_data
 }
 
+pub static BLOCKS: Blocks = get_block_data();
```
benjamin051000: This is where all the blocks live now. This is what you `use` in other files.

src/data_processing.rs
```diff
@@ -141,7 +141,7 @@ pub fn generate_world(
     let total_iterations_grnd: f64 = (scale_factor_x + 1.0) * (scale_factor_z + 1.0);
     let progress_increment_grnd: f64 = 30.0 / total_iterations_grnd;
 
-    let groundlayer_block = if args.winter { SNOW_BLOCK } else { GRASS_BLOCK };
+    let groundlayer_block = if args.winter { BLOCKS.by_name("snow_block").unwrap() } else { BLOCKS.by_name("grass_block").unwrap() };
```
benjamin051000: This is... not the most elegant API. 🤷🏼‍♂️

benjamin051000:
This is a good idea, but right now the implementation doesn't hold up, for a few reasons:

- `toml::from_str` isn't const (because we don't yet have const traits, apparently)
  - This _can_ be resolved with `static-toml` crate
- The API is not as good as it used to be due to all the `unwrap`s and stuff 
  - Maybe a lot of it can be removed? I'm not great at rust so there may be a better way

benjamin051000:
>     * `toml::from_str` isn't const (because we don't yet have const traits, apparently)
Meh, I guess the performance penalty is small...


>       * This _can_ be resolved with `static-toml` crate
True. That might be worth looking into. It still requires a global variable, though, and it seems clunky and tough to massage your struct into something nice.

>     * The API is not as good as it used to be due to all the `unwrap`s and stuff
Returning an Option is probably the right thing to do. You can always just `?` it

src/block_definitions.rs
benjamin051000: This works because every entry in `blocks.toml` is called `[[blocks]]` right now. That's how this works. Okay

src/block_definitions.rs
benjamin051000: This no longer exists

benjamin051000:
Would it be better to use something like `static-toml` and keep the global var? Or should we just load this data into a var in `main()` and pass it around to functions? I'm not sure what's better.

If we had `const fn toml::from_str`, I'd simply use that and throw it into a static... I wonder if nightly rust supports this...

benjamin051000:
The other thing we can do (I think) is just fill the toml file with null values in every entry. It's a horrible thing to do, but it'll give us a much cleaner generated struct. Then we can have it be a `const` and get rid of all the unwrap logic. The price is a toml file with like 500 empty entries, though. 🤦🏼

benjamin051000:
For the time being, honestly maybe we should just unwrap, and get rid of it later when static_toml supports what we need it to support

louis-e:
Hey man, sorry again for the delay! Just resolved some merge conflicts so we can finally get this merged hah. Will let this run through the CI, I probably didn't catch everything at my first commit haha

louis-e:
I'm really sorry, I'm afraid I need your help here hahha! I already tried a few times locally rebasing your PR to get it up to date with the main branch, but I fail miserably. I can offer you to freeze any integration in the main branch until we get this integrated. Let me know if there are certain files I can help you with to get this done asap :D

benjamin051000:
You're alive, haha!

This PR is not ready for merging. I will sync it up to main. But note that my last commit was "down to 7 errors". What I meant by that comment is that it currently isn't building.

I wasn't even sure if you wanted this feature or not, but it seems like you do, so I'll get it synced with main

louis-e:
No worries, take your time. I like the approach of cleaning up the block definitions which is why I'm so tempted to get this running :D

We just need to make sure that the performance stays roughly the same and maybe find a solution for the unwrapping stuff. So if you think it's not so critical to get it into the next release, we can take our time for it. I just didn't want to leave you hanging for so long! Thanks a lot for all your effort! :)

benjamin051000:
closing this for now as I am not working on Arnis at the moment. Feel free to grab my branch and get it past the finish line if you wish.

PR #354, opened 2025-01-29
Add a XDG desktop file

TheComputerGuy96:
This allows proper packaging of the application for Linux distros

### DIFF
--- a/gui-src/arnis.desktop
+++ b/gui-src/arnis.desktop
@@ -1,0 +1,10 @@
+[Desktop Entry]
+Type=Application
+Name=Arnis
+Comment=Generate cities from real life in Minecraft
+Comment[lt]=Sugeneruokite tikrovės miestus „Minecraft“ žaidime
+Icon=arnis
+Exec=arnis
+Categories=Game;Utility;Geoscience
+Keywords=OSM;OpenStreetMap;Tauri
+StartupWMClass=arnis


gui-src/arnis.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Type=Application
+Name=Arnis
+Comment=Generate cities from real life in Minecraft
+Comment[lt]=Sugeneruokite tikrovės miestus „Minecraft“ žaidime
+Icon=arnis
```
benjamin051000, line 6: I don't think `arnis` is an Icon. But there are icons in https://github.com/louis-e/arnis/blob/main/icons/, maybe you could specify one of those?

gui-src/arnis.desktop
```diff
@@ -0,0 +1,10 @@
...
+Exec=arnis
+Categories=Game;Utility;Geoscience
+Keywords=OSM;OpenStreetMap;Tauri
```
benjamin051000, line 9: What do these keywords do? I see this in the XDG Desktop Entry Specification:

>  A list of strings which may be used in addition to other metadata to describe this entry. This can be useful e.g. to facilitate searching through entries. The values are not meant for display, and should not be redundant with the values of Name or GenericName. 

Does this provide additional keywords to discover the app when using the OS's search tool (e.g., KRunner or GNOME search)?

If I understand what this does correctly, I don't think arnis should show up if someone searches "Tauri" in their OS search bar, right? It's akin to adding "rust" to this field. I guess "OSM"/"OpenStreetMap" are okay. What about "Minecraft"?

gui-src/arnis.desktop
benjamin051000: Yeah, when running `desktop-file-install arnis.desktop`, I don't see the icon.

benjamin051000 (changes requested):
I ran `desktop-file-validate`, which I found from the [Arch Wiki](https://wiki.archlinux.org/title/Desktop_entries#Validation):

```
❯ desktop-file-validate arnis.desktop 
arnis.desktop: hint: value "Game;Utility;Geoscience" for key "Categories" in group "Desktop Entry" contains more than one main category; application might appear more than once in the application menu
arnis.desktop: hint: value item "Geoscience" in key "Categories" in group "Desktop Entry" can be extended with another category among the following categories: Education, or Science
```

Please fix the icon and see other comments (may or may not require changes)

gui-src/arnis.desktop
TheComputerGuy96: The .desktop file is intended as an alternative to the `tauri bundle` workflow (so the packager/advanced user can definitely put the icon in the correct place)

gui-src/arnis.desktop
TheComputerGuy96: > Does this provide additional keywords to discover the app when using the OS's search tool

I think that's the main reason for that option

louis-e:
Since there hasn't been any more commits to resolve the discussion, I will merge this and follow up with an own commit to fix the open points

PR #374, opened 2025-02-19
Set `terrain` arg to store true.

benjamin051000:
This was erroneously set to store false, which means it's impossible to enable the terrain from the cli.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -107,3 +107,57 @@ fn parse_duration(arg: &str) -> Result<std::time::Duration, std::num::ParseIntEr
     let seconds = arg.parse()?;
     Ok(std::time::Duration::from_secs(seconds))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_flags() {
+        // Test that winter/terrain/debug are SetTrue
+        let cmd = [
+            "arnis",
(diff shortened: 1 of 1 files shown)


benjamin051000:
Hm, this is failing with the following:
```
cargo r --no-default-features --release -- --bbox "-71.035194,42.327427,-71.008286,42.342210" --path "/home/benjamin/.var/app/org.prismlauncher.PrismLauncher/data/PrismLauncher/instances/Fabulously Optimized(3)/.minecraft/saves/testing" --terrain
```

(with a `dbg!` around the first line of `ground::fetch_elevation_data`) Prints:
```
[src/ground.rs:130:32] bbox_str.split_whitespace().map(|s: &str|
s.parse::<f64>()).collect::<Result<Vec<f64>, _>>() = Err(
    ParseFloatError {
        kind: Invalid,
    },
)
Warning: Failed to fetch elevation data: invalid float literal
```

benjamin051000:
I'll add a unit test that verifies this before merging.

benjamin051000:
Blocked on #378

benjamin051000:
@louis-e once CI passes, this is ready for signoff

louis-e:
So I guess we're starting with unit tests now ;)
Thanks a lot :)

PR #376, opened 2025-02-20
ci: run cargo test

benjamin051000:
In the future, we will start adding more and more unit tests (or at least, *I* will. 😉)

Time to start ensuring they pass in PRs


benjamin051000:
This should be merged after #374. The tests run really fast. 

If it has to rebuild, then maybe we remove the `cargo build` step since `cargo test` sorta runs `cargo build` first? 🤷🏼‍♂️

Issue #29, opened 2024-05-02
Suggestion: Co-ordinate sync with reality [OPT-IN]


goteusz-maszyk:
Instead of starting at 0,0 the generator would start at the position that would allow generating a few tows next to each other, in correct geographical positions. As many places would end up far away, It would log the position when finished

daniil2327:
Just generate the whole Earth!

louis-e:
I like the idea of mapping the real coordinates to Minecraft coordinates, I already added it to the README ToDo section. What type of implementation would you prefer? A hosted website for an easy coordinate conversion (including an easy bbox finder) or extending the command line tool with a conversion feature?

scd31:
> What type of implementation would you prefer?

I've been thinking about this a bit and I think the best way would be to encode metadata in the level.dat file, via some additional NBT tags. Then a CLI could read that metadata when doing scaling. It would eliminate the need to have the user input world details manually into the CLI (or into a web utility).

Semi-related, but eventually I'd like to write a tile server, similar to Dynmap, but optimized for Arnis. One of the things I'd like for it to be able to do is read this same metadata and present it in the web UI (possibly with a converter from real lat/lon to MC coords)

cherbert:
As great as this tool is it's pretty useless without this feature being a priority right now. Every time you go back and generate a new area it renders it at 0,0,0. It makes it impossible to generate large worlds.

amir16yp:
i would like to add, that you can use journeymap mod to render all the chunks (i think it calls it "automap" or something like that) and then if you are in creative mode you can just right click anywhere on the map and teleport

cherbert:
> i would like to add, that you can use journeymap mod to render all the chunks (i think it calls it "automap" or something like that) and then if you are in creative mode you can just right click anywhere on the map and teleport. this can serve as a temporary solution for when you need to get to somewhere specific and you dont know where it is

Not sure how this is relevant to this conversation?

amir16yp:
> > i would like to add, that you can use journeymap mod to render all the chunks (i think it calls it "automap" or something like that) and then if you are in creative mode you can just right click anywhere on the map and teleport. this can serve as a temporary solution for when you need to get to somewhere specific and you dont know where it is
> 
> Not sure how this is relevant to this conversation?

OP is asking about coordinate sync i.e a way to convert IRL coordinates into minecraft ones. im just saying if you need to get somewhere specific in the map (which is the use case) you can use journeymap. that is all

[...]

PR #377, opened 2025-02-20
Formal BBox struct

### LINKED ISSUE #375: Issue with parsing bbox coords in elevation code
While working on #374 , I found this issue in the elevation code:

https://github.com/louis-e/arnis/blob/a969d65ae4ab55717a32f2c9c1bf666452e374d6/src/ground.rs#L130-L133

Here, `split_whitespace` is the wrong method to call. In CLI mode, it fails, leading the elevation code to be skipped.

Does the GUI pass the bbox coords in with space delimiters? The CLI breaks unless you use comma delimiters.

We should settle on one way to do this, or even better, replace the bbox string with an actual struct so we don't have to parse it in worldgen logic.

benjamin051000:
We need a more standardized way to represent bounding boxes, as we kind of just throw the string around and parse it into the four corresponding values when we need to.

Fixes #375 

# To Do
- [x] Get GUI working with this 
- [x] Fix compile errors
- [x] Fix render bug

### DIFF
--- a/gui-src/js/main.js
+++ b/gui-src/js/main.js
@@ -313,8 +313,8 @@ function handleBboxInput() {
               lat2 >= -90 && lat2 <= 90 &&
               lng2 >= -180 && lng2 <= 180
           ) {
-              // Input is valid; trigger the event
-              const bboxText = `${lat1} ${lng1} ${lat2} ${lng2}`;
+              // Input is valid; trigger the event with consistent comma-separated format
+              const bboxText = `${lat1},${lng1},${lat2},${lng2}`;
               window.dispatchEvent(new MessageEvent('message', { data: { bboxText } }));
 
               // Show custom bbox on the map
--- a/src/bbox.rs
+++ b/src/bbox.rs
@@ -1,0 +1,101 @@
+use crate::geo_coord::GeoCoord;
+
+/// A checked Bounding Box.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct BBox {
+    /// The "bottom-left" vertex of the rectangle
+    min: GeoCoord,
+
+    /// The "top-right" vertex of the rectangle
+    max: GeoCoord,
+}
+
+impl BBox {
+    pub fn new(min_lat: f64, min_lng: f64, max_lat: f64, max_lng: f64) -> Result<Self, String> {
+        let vals_in_order = min_lng < max_lng && min_lat < max_lat;
+
+        if !vals_in_order {
+            return Err("Invalid BBox".to_string());
+        }
+
+        let min = GeoCoord::new(min_lat, min_lng)?;
+        let max = GeoCoord::new(max_lat, max_lng)?;
+
+        Ok(Self { min, max })
+    }
+
+    pub fn from_str(s: &str) -> Result<Self, String> {
+        let [min_lat, min_lng, max_lat, max_lng]: [f64; 4] = s
+            .split([',', ' '])
+            .map(|e| e.parse().unwrap())
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap();
+
+        // So, the GUI does Lat/Lng and no GDAL (comma-sep values), which is the exact opposite of
+        // what bboxfinder.com does. :facepalm: (bboxfinder is wrong here: Lat comes first!)
+        // DO NOT MODIFY THIS! It's correct. The CLI/GUI is passing you the numbers incorrectly.
+        Self::new(min_lat, min_lng, max_lat, max_lng)
+    }
+
+    pub fn min(&self) -> GeoCoord {
+        self.min
+    }
+
+    pub fn max(&self) -> GeoCoord {
+        self.max
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_valid_input() {
+        assert!(BBox::new(0., 0., 1., 1.).is_ok());
+
+        assert!(BBox::new(1., 2., 3., 4.).is_ok());
+
+        // Arnis, Germany
+        assert!(BBox::new(9.927928, 54.627053, 9.937563, 54.634902).is_ok());
+    }
+
+    #[test]
+    fn test_from_str_commas() {
+        const ARNIS_STR: &str = "9.927928,54.627053,9.937563,54.634902";
+
+        let bbox_result = BBox::from_str(ARNIS_STR);
+        assert!(bbox_result.is_ok());
+
+        let arnis_correct: BBox = BBox {
+            min: GeoCoord::new(9.927928, 54.627053).unwrap(),
+            max: GeoCoord::new(9.937563, 54.634902).unwrap(),
+        };
+
+        assert_eq!(bbox_result.unwrap(), arnis_correct);
+    }
+
+    #[test]
+    fn test_from_str_spaces() {
+        const ARNIS_SPACE_STR: &str = "9.927928 54.627053 9.937563 54.634902";
+
+        let bbox_result = BBox::from_str(ARNIS_SPACE_STR);
+        assert!(bbox_result.is_ok());
+
+        let arnis_correct: BBox = BBox {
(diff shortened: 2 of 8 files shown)


benjamin051000:
Blocked on #378

benjamin051000:
Okay, on my local machine, I've definitely swapped a lat/lng somewhere... because renders are no longer working :( I'll try to figure out which

src/bbox.rs
```diff
@@ -0,0 +1,58 @@
...
+            });
+        }
+
+        Err("Invalid BBox".to_string())
+    }
+
+    pub fn from_str(s: &str) -> Result<Self, String> {
+        let [min_lat, min_lng, max_lat, max_lng]: [f64; 4] = s
+            .split(',')
+            .map(|e| e.parse().unwrap())
```
benjamin051000, line 30: Should this return an `Err` signifying parsing failure?

src/bbox.rs
```diff
@@ -0,0 +1,58 @@
...
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap();
```
benjamin051000, line 33: Should this return an `Err`? What cases will the `try_into` fail?

src/main.rs
```diff
@@ -401,26 +392,13 @@ fn gui_start_generation(
...
     terrain_enabled: bool,
 ) -> Result<(), String> {
+    use bbox::BBox;
+
     tauri::async_runtime::spawn(async move {
         if let Err(e) = tokio::task::spawn_blocking(move || {
-            // Utility function to reorder bounding box coordinates
-            fn reorder_bbox(bbox: &[f64]) -> (f64, f64, f64, f64) {
-                (bbox[1], bbox[0], bbox[3], bbox[2])
-            }
```
benjamin051000, line 414: hm, do I still need to do this? Maybe we can just make the GUI return it in the right order in the first place...

src/retrieve_data.rs
```diff
@@ -130,7 +131,10 @@ pub fn fetch_data(
     .relsinbbox out body;
     .waysinbbox out body;
     .nodesinbbox out skel qt;"#,
-        bbox.1, bbox.0, bbox.3, bbox.2
+        bbox.min().lng,
+        bbox.min().lat,
+        bbox.max().lng,
+        bbox.max().lat
```
benjamin051000: I'm not rendering buildings anymore... It's gotta be this that's causing the issue.

benjamin051000: Okay, I might be wrong, because when I switch these around, the API doesn't return any data. I guess I'm stumped as to why this no longer works.

benjamin051000: image

It's rendering the grass, but nothing on top of it.

benjamin051000: Okay, it looks like the issue was the order I was constructing... see https://github.com/louis-e/arnis/compare/f705619aab218cadc7e370c663f5f95f7117a147..cbf8c53312e2846e08e945d905ee906b9fd46ac4 for details. I think it's fixed!

benjamin051000:
This may or may not work with the GUI. I don't know, because I don't build or test the GUI. Soooo yeah, do with that knowledge what you will.

louis-e:
My bbox code was a mess hahha! It doesn't run in GUI mode yet, but I'll take care of that and merge it!

louis-e:
Alright, it's 3:30am now - let me know if this looks good to you or if I messed something up in your code! :D

XianlinSheng:
Does it sound good to represent the position data in two spaces: one real world system (in maybe 'spherical.rs' using LLPoint(the GeoCoord here) and LLBBox (the BBox here)), and a transformed mc system (in cartesian.rs currently, using XZPoint and XZBBox). This will make coordinate transformation much flexible and clearer. I just created the XZBBox struct and based the world editor on XZPoint, XZBBox, so it does not have to generate at (0,0), and I reserved the possibility to extend for more complex bbox shapes.

louis-e:
That sounds interesting, thanks for your idea. Can we merge this PR and follow up with your suggestion in https://github.com/louis-e/arnis/pull/409/ or a seperate PR? It would be cool to cover https://github.com/louis-e/arnis/issues/29 with this and enable multiple generations in a single world, since we can also programmatically set the player spawn point.

XianlinSheng:
Sure. You can merge this PR first and I will keep my #409 updated. These two are the components in a bigger architecture, and I will try to make them work seamlessly.

src/bbox.rs
```diff
@@ -25,13 +25,22 @@ impl BBox {
...
+        
+        // Parse the four floating point values
+        let mut values = [0.0; 4];
+        for (i, part) in parts.iter().enumerate() {
+            values[i] = part.parse::<f64>().map_err(|e| {
+                format!("Failed to parse coordinate value '{}': {}", part, e)
+            })?;
+        }
+        
+        let [min_lat, min_lng, max_lat, max_lng] = values;
```
benjamin051000: I'd like to go back to the old logic if possible. 

Can we just do one of these:
-  ~~`.split([',', ' '])` for both chars (but do we really need to support both space and comma-delim values? Maybe we should just pass the delimiter into the constructor here.~~ I see that 1d31dfe fixed this
- IIRC `.try_into()` will supply a good-enough error that it was unable to convert into an array of 4 floats.
- The new float-parsing logic adds several lines.

src/bbox.rs
```diff
@@ -70,6 +79,21 @@ mod tests {
 
         assert_eq!(bbox_result.unwrap(), arnis_correct);
     }
+    
+    #[test]
+    fn test_from_str_spaces() {
```
benjamin051000, line 80: If we _need_ space-delimited strings for GUI purposes, that's fine, but it would be ideal if the CLI and GUI used the same convention.

src/bbox.rs
```diff
@@ -43,8 +43,8 @@ impl BBox {
                 .map_err(|e| format!("Failed to parse coordinate value '{}': {}", part, e))?;
         }
 
-        let [min_lat, min_lng, max_lat, max_lng] = values;
-        Self::new(min_lat, min_lng, max_lat, max_lng)
+        let [min_lng, min_lat, max_lng, max_lat] = values;
+        Self::new(min_lng, min_lat, max_lng, max_lat)
```
benjamin051000: Hmm, I thought I had this right. What caused the need to change this? I definitely could have been wrong before, but we should make sure we understand what the ground truth is and probably write it down somewhere so there isn't any more confusion 😅

benjamin051000 (commented):
I'm not able to request changes against my own PR, interesting. Well this is what I'm seeing. 

My understanding is that we want this merged soon. I think I'll reset back to the first commit, rebase off main, and then reapply relevant changes to resolve the issues I have with the code changes.

gui-src/js/main.js
```diff
@@ -313,8 +313,8 @@ function handleBboxInput() {
               lat2 >= -90 && lat2 <= 90 &&
               lng2 >= -180 && lng2 <= 180
           ) {
-              // Input is valid; trigger the event
-              const bboxText = `${lat1} ${lng1} ${lat2} ${lng2}`;
+              // Input is valid; trigger the event with consistent comma-separated format
+              const bboxText = `${lat1},${lng1},${lat2},${lng2}`;
```
benjamin051000, line 317: If we do this, then we shouldn't need space-delimited strings in the Bbox constructor, right?

benjamin051000:
Made some updates. Moving back to draft until it's working again

src/bbox.rs
benjamin051000: nah, if it fails we will certainly not be able to move on. Just panic now

src/bbox.rs
benjamin051000: no just panic

src/bbox.rs
benjamin051000: Right now it does both. I'll bring back this unit test

benjamin051000:
@louis-e ready for review, please see the comments I wrote

louis-e:
Thanks a lot for your work, looks good to me!! :)

PR #392, opened 2025-03-02
fix grass in a residential area

V14es:
Untitled


src/element_processing/landuse.rs
```diff
@@ -75,6 +75,8 @@ pub fn generate_landuse(
             editor.set_block(block_type, x, ground_level + 1, z, None, None);
         } else if landuse_tag == "construction" || landuse_tag == "railway" {
```
benjamin051000, line 76: Would it be better to combine the "grass" check with this line?

PR #403, opened 2025-04-05
Some improvements for the generation

Oleg4260:
Changes: 
- Increased `MAX_Y` to 319 because it's the new height limit of Minecraft
- Added more entries for the `natural` tag, like bare rock, glaciers and others
- Removed trees from farmlands because trees don't grow randomly in the middle of land that is used for growing crops
- Improved winter mode by replacing water with ice (except rivers) and removing crops from farmlands
- Fixed vertical scaling so that it now matches horizontal scaling
- Replaced 15x15 water grid in farmlands with singular water blocks every 8 blocks
- Reworked generation of upper dirt layer, now there's no dirt under stone, sand and ice
- Removed randomly scattered benches (3 oak log blocks) from parks

### DIFF
--- a/src/ground.rs
+++ b/src/ground.rs
@@ -5,8 +5,7 @@ use image::{Rgb, RgbImage};
 const MIN_Y: i32 = -62;
 /// Maximum Y coordinate in Minecraft (build height limit)
 const MAX_Y: i32 = 319;
-/// Scale factor for converting real elevation to Minecraft heights
-const BASE_HEIGHT_SCALE: f64 = 1.0;
+
 /// Mapbox API access token for terrain data
 const MAPBOX_PUBKEY: &str =
     "pk.eyJ1IjoiY3Vnb3MiLCJhIjoiY2p4Nm43MzA3MDFmZDQwcGxsMjB4Z3hnNiJ9.SQbnMASwdqZe6G4n6OMvVw";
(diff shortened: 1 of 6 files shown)


Oleg4260:
Trying new dot pattern instead of the old 15x15 grid. I think this is better because less water is seen, also the old pattern didn't irrigate all the blocks

src/ground.rs
```diff
@@ -4,9 +4,9 @@ use image::{Rgb, RgbImage};
 /// Minimum Y coordinate in Minecraft (bedrock level)
 const MIN_Y: i32 = -62;
 /// Maximum Y coordinate in Minecraft (build height limit)
-const MAX_Y: i32 = 256;
+const MAX_Y: i32 = 319;
 /// Scale factor for converting real elevation to Minecraft heights
-const BASE_HEIGHT_SCALE: f64 = 0.6;
+const BASE_HEIGHT_SCALE: f64 = 1.0;
```
benjamin051000: This variable should probably be removed if it's a `const` that is a factor of 1.

benjamin051000 (approved):

src/ground.rs
Oleg4260: Good point, I just made a new commit removing that const, also I decided to remove that .sqrt() that mismatches vertical and horizontal scaling when you change the scale factor in the settings.

PR #404, opened 2025-04-09
Choose where you spawn on the map

Geotree12:
First time contributing to a code base so sorry if it's not too good right now, had to make a couple helper functions from osm_parser.rs public to translate from lng/lat coordinates to minecraft coordinates as well.

New parameter is added 
--spawn-point="longitude,latitude"

when this parameter is not specified the spawn point defaults to 0,0


Oleg4260:
Looks good, but it doesn't pass the checks because of bad formatting, you should execute `cargo fmt` and push another commit to fix this

Geotree12:
got it all fixed up now, clippy was a bit cranky but it should all be good.

louis-e:
I'll wait for https://github.com/louis-e/arnis/pull/409 to be merged so that we can also include a GUI option for it / set the spawn point automatically!

louis-e:
I reworked your code into the current code base with https://github.com/louis-e/arnis/pull/461 since a lot has changed in the architecture. So thanks for your contribution :)

Issue #399, opened 2025-03-29
[BUG] Generated landscape does not match the real landscape


FreeScissors311:
**Describe the bug**
The landscape of the generated map does not match the landscape in OpenStreetMap.
**Used bbox area**
53.271168 69.333858 53.296004 69.376602

**Arnis and Minecraft version**
Arnis: v2.2.0
Minecraft: v1.21.1

**Screenshots**
![Image](https://github.com/user-attachments/assets/e6ab8794-1153-4dd1-9d63-d778ec59504d)

![Image](https://github.com/user-attachments/assets/ab5d04fd-5e44-4dc4-a178-7048c327b09a)

**Additional context**
The objects that are marked in the screenshot are not at all where they should be in Minecraft.

Oleg4260:
Looks similar to [this](https://github.com/louis-e/arnis/issues/325#issuecomment-2613578083)

louis-e:
Fixed! I've resolved the building offset issues when terrain mode is enabled. Buildings should now properly align with the landscape and maintain their correct positions relative to the terrain elevation. The fix will be included in the next release!

Issue #325, opened 2025-01-22
[DISCUSSION] Elevation Feedback!


louis-e:
Hi there! With the recent release, elevation was implemented into the generated Minecraft world. We'd love to hear your feedback if you like it, on how it looks, its performance and if there are any issues you have encountered! :)
Also feel free to share screenshots here if you encounter any nice locations!

Simpsons4ever:
I know about soviet buildings very well, let's take with this 1960s apartment:
Я очень хорошо знаю насчёт советских зданий, давайте возьмём пример с квартиры 1960-ых годов:

![Image](https://github.com/user-attachments/assets/d94473d1-ff94-4658-a642-f0e53378db2e)

![Image](https://github.com/user-attachments/assets/fbe9701b-9801-4762-98c9-62d9fa1fc746)
And remember, if player doesn't fit in, just replace roof material. Instead of block, we replace with slab, it's simple
И запомните, если игрок не влезает, надо заменить материал крыши. Вместо блоков, надо заменить плитками, это всё просто

Simpsons4ever:
@Oleg4260 Смотря на то, в какую серию входит каждая квартира

mpr90:
Very cool feature, thanks. Been trying out a few areas, and unfortunately the terrain doesn't seem properly aligned with the other map features (roads, buildings, etc). It looks like there is an offset. For example, when using the bbox 37.976173 -122.575922 37.985001 -122.560258, you can see from the screenshots that in google earth (and reality), the large hill is behind the car park and supermarket. However, in Arnis/minecraft, the car park is flowing up the hill. This is just one example...

<img width="1380" alt="Image" src="https://github.com/user-attachments/assets/60bd8480-041e-4f4f-9eb5-e03032ea7eed" />

![Image](https://github.com/user-attachments/assets/cae3e5d9-7bc1-4e00-b9d5-8e8344e24a40)

Oleg4260:
@mpr90 I agree, I noticed the same problem. For example. the lake here isn't aligned with the bottom of the quarry

![Image](https://github.com/user-attachments/assets/7c9715a9-83ee-4ded-937e-b08dff950596)

![Image](https://github.com/user-attachments/assets/9f92705c-19a3-4cfd-8538-70817520dc52)

deadstalin78:
![Image](https://github.com/user-attachments/assets/e7c9d7c0-ab92-4665-a725-b1d9e6769e87)

[...]

PR #406, opened 2025-04-10
Minor adaptions for PR 403

louis-e:


src/data_processing.rs
```diff
@@ -170,7 +170,6 @@ pub fn generate_world(
                 if !editor.check_for_block(x, max_y, z, Some(&[STONE]), None) {
                     editor.set_block(groundlayer_block, x, max_y, z, None, None);
                     editor.set_block(DIRT, x, max_y - 1, z, None, None);
-                    editor.set_block(DIRT, x, max_y - 2, z, None, None);
```
Oleg4260, line 173: This shouldn't be removed, it looks much better with 2 blocks of dirt under it, like in a normal minecraft world, also this depth might be needed in the future for some features that go in the ground

src/element_processing/landuse.rs
Oleg4260, line 1: I'm sorry, but what you did is really weird. You basically cancelled almost everything I did, you made trees randomly spawn again and the ice in the winter mode now looks weird, also this code is much bigger and more complicated. Could you please revert this file back to the original state as it was when I commited it? it looked much better.

image

src/ground.rs
Oleg4260, line 1: Do you keep this to prevent very steep differences in elevation? I don't like that scale.sqrt() because it distorts the scale even more. Also when there wasn't BASE_HEIGHT_SCALE the hills had their real size

Oleg4260:
Hi Louis, thanks for merging my PR, but looks like we have some disagreement here, and you cancelled a lot of my changes and made some parts worse. Please read all the comments to the code, I want to ask you to revert some of your changes in this commit

src/data_processing.rs
louis-e: I removed it for generation speed reasons, but I understand your point. I've noticed some generation performance issues since the last release and wanted to counteract this by removing this layer for now until we managed to improve the performance again. I can add the line again in the meanwhile

src/ground.rs
louis-e: With the addition of elevation we have the problem that some large buildings on bigger elevation differences basically "merge with the ground". I attached an image to demonstrate what I mean.
image

When testing your PR I've seen this happen on the first generation already, which is why I reverted it. Do you have an idea on how we can counteract this?

src/element_processing/landuse.rs
louis-e: I tend to generate rural areas during development and in my personal opinion it looked better with a bit of random tree generation on the fields. I'll look into this again with some areas 

Regarding the ice generation in winter; I was thinking it looked a bit odd so I tried to give it some more randomness. I agree, this doesn't look good either - I didn't test this one thoroughly.

louis-e (commented):
Thanks a lot for your feedback @Oleg4260, let me hear your thoughts on my comments and then we can proceed with the next steps!

src/ground.rs
Oleg4260: Ok, I've noticed that as well, I think this can be fixed by adding generation of the basement floor, so that the first floor is elevated above the ground, after fixing this issue we can increase the elevation scale to 1.0

Many buildings have a basement and their first floor is elevated about 1 meter above the ground, so this would look good
image

Oleg4260: I imagine it to look like this. The first floor of the building has to start on the highest point and have a basement filling the gap under it.
image

Oleg4260: Also please note the issue of OSM data and elevation data being not aligned perfectly, it is noticeable in many generations, see #399

src/element_processing/landuse.rs
Oleg4260: I understand, with trees the farmlands look a little more detailed, but they are meant to be smooth, no farmer would be happy to see so many trees taking space in their field and not letting tractors move. Speaking of the winter mode, neither mine symmetric option nor your random look good, but I think mine could be there temporarily, its code is more compact and probably more fast, so I'm asking you again to revert this file to my version if you agree with me

image

src/element_processing/leisure.rs
```diff
@@ -112,7 +112,13 @@ pub fn generate_leisure(
                     let random_choice: i32 = rng.gen_range(0..1000);
 
                     match random_choice {
-                        0..40 => {
+                        0 => {
+                            // Benches
+                            editor.set_block(OAK_LOG, x, ground_level + 1, z, None, None);
+                            editor.set_block(OAK_LOG, x + 1, ground_level + 1, z, None, None);
+                            editor.set_block(OAK_LOG, x - 1, ground_level + 1, z, None, None);
+                        }
```
Oleg4260, line 120: I have another question about these benches. There are too many of them, they are scattered randomly and they don't look good. Also the program already has the code to generate benches that are mapped in the OSM, so why would you not want to remove this part of code?

Oleg4260:
Thanks for your reply and all the explanations, @louis-e, just tagging you in case you didn't get notified about my answers

louis-e:
Hey man, thanks for the ping - I didn't notice your updates. Will look into it on the weekend, thanks for all your input! :)

src/ground.rs
louis-e: I see, that could be an idea. It'd be good to have a POC of this so we can take a look at it visually. If I have the time I can try to do that and ping you with the results.
Also of course the current elevation 'stretching' doesn't get rid of this issue either, but I observed it way less. That's why I'd keep it as it is for now - but we should definitely reconsider it before creating the next release!

src/element_processing/landuse.rs
louis-e: Thanks for the great discussion, I just reverted it to your change!

src/element_processing/leisure.rs
louis-e: Pretty much the same thing here as with the trees on farmland - I originally added the generation on leisure because it added a bit more variety. I'll revert the lines as well, maybe we can come up with a good looking bench structure soon hah!

src/ground.rs
louis-e: Before and after
image
image

Oleg4260: This really looks good. But I wonder if the left door is meant to go into the basement? This building might not be meant to be located on a side of a hill, but on the top, since the issue of OSM data and elevation data is still present so it also should be fixed, as I said before, it's in #399 and also [here](https://github.com/louis-e/arnis/issues/325#issuecomment-2613578083). As I noticed, on different hemispheres of the earth the terrain is shifted into a different direction, do you have any idea how to align terrain with osm objects properly?

louis-e: Thanks for your feedback, do you think the foundation should be a kind of stone block for all buildings or should it be the wall block of the building? For the OSM and elevation data shift - that's going to be a tricky task to fix haha! Not sure yet what causes this, I wanted to look into this with a bit of more time. Also thanks for the hint with the hemispheres, I wasn't aware of that yet.

Oleg4260: I think it should be the same as the wall block for now

Issue #274, opened 2025-01-12
[FEATURE] Snap roads and features to 90° 


mattdm:
**Is your feature request related to a problem? Please describe.**

This is related to #97, and probably also to #102. As the issue about map rotation notes, many cities are not aligned to the compass points. However, it's even messier than that -- many older cities have sections with different overall alignments, or maybe even no plan at all. This results in a lot of jagged roads no matter the overall rotation.

**Describe the solution you'd like**F

Roads which are some amount skewed from alignment should be snapped to that alignment. That is, if a road runs at 93°, alter it to run at 90° instead.

There might be two configurable options:

1. snap if difference is less than N°
2. snap to
    * 90° alignments only
    * also include 45°f diagonals
    * snap to closest 15°

**Additional context**

I  know this is a _lot_ more complicated than the current approach!

briward:
This sounds potentially hard to do but a really cool feature if it were possible.

XianlinSheng:
I am working on this. I have tested a simple iteration algorithm, and it should not be too difficult to accomplish.

Issue #97, opened 2024-12-29
[FEATURE] The ability to rotate maps


faithsapling:
The streets of most cities don't go perfectly east to west or north to south, they tend to slant sideways. As seen in this image, when you create a rectangle, it's inevitable that you'll just be getting a diagonal section of a city. It would be nice if we could have maps rotated to solve this problem. 

![Screenshot 2024-12-29 082705](https://github.com/user-attachments/assets/d50d43c9-9e95-4a60-a372-65db034ff9b1)

louis-e:
That's a very good feature request, thanks! I will look into this and use this issue as an update thread for this feature :)

omrischwarz:
There are quite a few American cities with Cartesian street grids that are not aligned with the cardinal directions, and the ability to rotate the Minecraft grid to match them would produce results that are as good as those that use Chicago. New York, Philadelphia, Boston, Minneapolis, and several others have such street grids.

Kuki77:
Hi,
I just tried Arnis for our small village (in Germany) and it has also the characteristics, that the streets are not N/S or S/W oriented and even if most houses / buildings are rectangular, arnis is making them quite weird. 
Before starting a new request I searched and found this one, very  good.

So please, if it's possible, add an angle parameter for rotating the map (easier way) or a slider to rotate the map before generation.

CapaParlo:
I don't know where to tell so will do here. It's a good project but you difinitely gotta add some terrain. Like mate, satellites know how high the ground is, same as the ground under the water. But yeah, you are limited by height. Well, at least 50% relief scale could be cool! And some things like big objects needs to be added. Thank you for idea and realization! Hope it would be 3d in some time

CapaParlo:
You can use this to add mountains under the buildings, it would be way recognizeable and comprehendable with landscape relief https://uk-ua.topographic-map.com
Just make some nodes to adjust how buildings would touch the ground so it would look properly

fish4terrisa-MSDSM:
Any updates on the rotate feature?
Many cities and universities unfortunately aren't designed for arnis and love to slant a lot...
It'll help so much for people using arnis to create cute models of their surrounding

InternetOfTofu:
I really want this feature too, so I spend a little time looking into how to potentially implement this feature. I didn't thoroughly read the codebase, so some of the following might be based on incorrect assumptions. I am also no GIS expert, and have little knowledge in this field. Please correct me if I got it wrong. 

Observation:
[...]

PR #409, opened 2025-04-13
Open up a map editing architecture in the workflow for transformation, and enables generating at non-origin

XianlinSheng:
1. Added a map_editing step in the workflow, after parsing element to ProcessedElement, before generating_world(...). This enables development for map transformation, e.g. translating (#29), rotating (#97), snapping to grid (#274). To define operations, edit map_editing.json, which is a list of operation config dict that will be deserialized to construct specific types of Operator. The operation is applied on i32 coordinate, and i suppose 1m resolution is accurate enough for a city.
2. Enabled world generation at non-origin, which is a necessary feature for map editing.
journeymap
remotesight


louis-e:
Thanks a lot for your effort! I'll take some time to look into this to make sure that I fully understand the changes and then follow up with merging it!

benjamin051000:
Whoaaaaaa this is cool. I'll do a code review when I finish BBox PR

src/coordinate_system/cartesian/xzbbox.rs
```diff
@@ -0,0 +1,84 @@
...
+        XZBBox {
+            point1: XZPoint { x: 0, z: 0 },
+            point2: XZPoint {
+                x: scale_factor_x as i32,
+                z: scale_factor_z as i32,
+            },
+        }
+    }
+
+    pub fn nblock(&self) -> u64 {
```
benjamin051000: I'm unsure of what `nblock` means. Number of blocks? I don't know what that means in this context. Can we please use full names for methods + comments describing how they work? Much appreciated 🙏🏼

src/coordinate_system/cartesian/xzbbox.rs
```diff
@@ -0,0 +1,84 @@
...
+        let nx = self.point2.x - self.point1.x + 1;
+        let nz = self.point2.z - self.point1.z + 1;
+
+        (nx as u64) * (nz as u64)
```
benjamin051000: This can probably just call `nblock_x` and `nblock_z` to reduce repetition. However I'm not sure what any of these functions actually do, so I can't say for sure if that's a sensible suggestion.

src/coordinate_system/cartesian/xzbbox.rs
```diff
@@ -0,0 +1,84 @@
+use super::xzpoint::XZPoint;
+use super::xzvector::XZVector;
+use serde::Deserialize;
+use std::ops::{Add, AddAssign, Sub, SubAssign};
+
+#[derive(Debug, Deserialize, Copy, Clone)]
+pub struct XZBBox {
+    pub point1: XZPoint,
+    pub point2: XZPoint,
+}
```
benjamin051000: I'm imagining this as a 2-dimentional box represented by two 2-dimensional points (because in minecraft Y is the height, so everything here is within a 2D plane).

Are there well-defined rules about the semantics of `point1` vs `point2`? For example, is one of them always on the "top-left"? 

I suppose the only "invariant" here is that the two corners must be opposite one another, as to define a Box. 

So the following must hold:
```rust
assert_ne!(point1.x, point2.x);
assert_ne!(point1.y, point2.y);
```
hmm... Unless it's not unusual to have a 1-dimensional Box?

---

We should discuss these points so that we can (1) better define what these vars do in the context of Arnis, and (2) ensure this struct is never misused by defining some invariants like the one above (see how I ensure BBox is never misused in #377 ).

src/coordinate_system/cartesian/xzbbox.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl XZBBox {
+    pub fn from_scale_factors(scale_factor_x: f64, scale_factor_z: f64) -> Self {
```
benjamin051000: Thank you for the full variable names. Makes this much easier to understand at a glance.

src/coordinate_system/cartesian/xzbbox.rs
```diff
@@ -0,0 +1,84 @@
...
+        nz as u32
+    }
+
+    #[inline]
+    pub fn contains(&self, xzpoint: &XZPoint) -> bool {
+        xzpoint.x >= self.point1.x
+            && xzpoint.x <= self.point2.x
+            && xzpoint.z >= self.point1.z
+            && xzpoint.z <= self.point2.z
+    }
```
benjamin051000: `contains` looks good.

I am wary about manually `#[inline]`ing stuff until I've seen proof via benchmarks that a performance issue exists here, but that's just a nit.

src/coordinate_system/cartesian/xzbbox.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// below are associated +- operators
+impl Add<XZVector> for XZBBox {
+    type Output = XZBBox;
```
benjamin051000: unused?

src/coordinate_system/cartesian/xzbbox.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    fn add(self, other: XZVector) -> XZBBox {
```
benjamin051000: This consumes the `other` vector. Is that what we want? Just making sure

benjamin051000: same on line 80

src/coordinate_system/cartesian/xzbbox.rs
benjamin051000: same on line 69

benjamin051000: Oh, maybe this is required to impl the trait? Sorry I'm a rust noob

src/coordinate_system/cartesian/xzvector.rs
```diff
@@ -0,0 +1,45 @@
+use serde::Deserialize;
+use std::ops::{Add, AddAssign, Sub, SubAssign};
+
+#[derive(Debug, Deserialize, Copy, Clone)]
+pub struct XZVector {
```
benjamin051000, line 7: This file looks very similar to XZPoint. Can they be combined?

src/coordinate_system/cartesian/xzbbox.rs
benjamin051000: I later discovered that this gives you the total number of "blocks" (or whatever unit this box uses) inside it, AKA the area. Let's rename this to make that more clear.

src/coordinate_system/cartesian/xzbbox.rs
benjamin051000: However, I'm not sure why I would want to make a Box from two scale factors. What does that even mean? 😕 

Are we making a box that has a scale? Or are we making a box used to scale other stuff? Not sure

src/map_editing/edit_map.rs
```diff
@@ -0,0 +1,40 @@
+use super::operator::Operator;
+use crate::coordinate_system::cartesian::XZBBox;
+use crate::osm_parser::ProcessedElement;
+use crate::progress::emit_gui_progress_update;
+use colored::Colorize;
+use std::fs;
+
+pub fn edit_map(elements: &mut Vec<ProcessedElement>, xzbbox: &mut XZBBox) {
```
benjamin051000: I prefer "transform map" as the term, "edit" is too unclear, since we "edit map" later by overwriting the minecraft save file

src/map_editing/operator.rs
```diff
@@ -0,0 +1,44 @@
...
+use crate::coordinate_system::cartesian::XZBBox;
+use crate::osm_parser::ProcessedElement;
+
+pub enum Operator {
+    Translate(Translator),
+}
+
+impl Operator {
+    pub fn operate(&self, elements: &mut Vec<ProcessedElement>, xzbbox: &mut XZBBox) {
+        match self {
```
benjamin051000: Hmmmm. I think this works best if Operator is a trait. That way you can just make `fn operate()` a trait method and `Translator` implements it. Then you just call `operate()` and it dispatches to the proper one.

I think? Unless you see a reason why that wouldn't work. What you've essentially made is dynamic dispatch with this match statement :D

src/map_editing/translate/translator.rs
```diff
@@ -0,0 +1,63 @@
+use super::startend_translator::StartEndTranslator;
+use super::vector_translator::VectorTranslator;
+use crate::coordinate_system::cartesian::{XZBBox, XZVector};
+use crate::osm_parser::ProcessedElement;
+
+// types of translation
+pub enum Translator {
+    Vector(VectorTranslator),
+    StartEnd(StartEndTranslator),
+}
```
benjamin051000: I'm not sure if these need to be their own structs. Maybe it's helpful. An alternative would just be 
```rust
pub enum Translator {
    Vector(XZVector),
    StartEnd(XZPoint, XZPoint),
}
```
I'm not sure if this is better or worse, but it removes 2 files. What are your thoughts? I could be totally wrong here

src/map_editing/edit_map.rs
```diff
@@ -0,0 +1,40 @@
...
+    println!("{} Editing data...", "[3/6]".bold());
+    emit_gui_progress_update(11.0, "Reading map editing config...");
+
+    match fs::read_to_string("map_editing.json") {
```
benjamin051000: This path needs to be a CLI argument (in this PR). Also probably added to the GUI (doesn't need to be part of this PR)

map_editing.json
```diff
@@ -0,0 +1,14 @@
...
+            "type": "vector",
+            "config": {
+                "vector": {
+                    "dx": 500,
+                    "dz": 500
+                }
+            }
+        }
+    }
+]
```
benjamin051000: Two things:

1. Any particular reason to use JSON over something else? I've set precedent to use TOML in #351 but that isn't merged. Also it doesn't really matter. Maybe JSON makes the most sense if you're sending it from the GUI which is a web-frontend. Okay, I'm cool with JSON. As long as the decision is well-thought out
2. This should be an example file (change the filename to `example_transformations.json`). Also it should show every possible thing you can do (like StartEnd?) as a form of documentation

benjamin051000 (changes requested):
Overall AWESOME idea and implementation. I've left many notes because I'm passionate about this change and want to see it perfected and merged! 

> [!NOTE] 
> I am a Rust noob. If some of these comments are wrong, feel free to tell me I'm wrong and hit resolve. No hard feelings :D 

# Overall thoughts
- I _like_ the Operator design. Very nice and flexible. How hard would it be to add things like repetition operators (e.g., repeat a transformation/rotation several times)? Looking forward to seeing how simple adding rotation will be :D 
- Please add unit tests wherever you see fit. I know they're pretty simple data structures, but it gives us additional confidence & debug info when weird edge cases arise.
- It would also be useful to be able to pass transformations via CLI arguments. Something like `--transform translate vector 500 500` or something. That may be complex and not that great though so we can discuss separately

I'm willing to approve this* if you review my comments for important suggestions first.

\* I am not the maintainer so my words have 0 weight

src/coordinate_system/cartesian/xzbbox.rs
benjamin051000: Update, I'm almost certain you can make this function return
```rust
(self.nblock_x(), self.nblock_z())
```

src/coordinate_system/cartesian/xzbbox.rs
benjamin051000: To be clear, the questions I need answered are:
- Under what circumstances is a XZBBox valid? Across all XZPoints? 
- Under what circumstances is an XZPoint valid? Across all integers? 

Knowing this will help us define unit tests & ensure we never accidentally create an invalid XZBBox/XZPoint, leading to less bugs :D

src/coordinate_system/cartesian/xzbbox.rs
benjamin051000: again, this may be required by the trait. If it's correct, I will go read up on these trait definitions

XianlinSheng:
Thanks a lot for your time. I will response to every comment thoroughly and give you my thoughts on some of the designs when I have enough time on this, maybe on this weekend.

src/coordinate_system/cartesian/xzbbox.rs
XianlinSheng: No problem to rename it to a clearer one, but I would like to use something like 'total_blocks' because 'area' sounds to me like a length squared dimensioned float physics property. 'total_blocks' will be consistent in style with total_blocks_x and ..._z. If it sounds good I will rename all related things.

src/coordinate_system/cartesian/xzbbox.rs
XianlinSheng: Good to notice this, I will improve these

src/coordinate_system/cartesian/xzbbox.rs
XianlinSheng: parse_osm_data(...) takes the bbox_tuple, etc., and outputs parsed_elements and these two scale factors. Originally the world is then generated from 0..scale_factor_x. I assume you have a similar question when I see these two variables. scale_factor is not scaling ratio, which is technically called world_scale in main.rs. 

I suppose it is called this because the global LL coordinates are first normalized so that bbox becomes 0..1. The normalized coordinates are then scaled up to distance in meter by these two factors. So these two scale factors means the xz lengths of generated world. 

These two will not be necessary in the future because when your LLBBox is done, we will call coordinate transformation functions to directly convert LLPoint to XZPoint.

src/coordinate_system/cartesian/xzbbox.rs
XianlinSheng: OK, you are more an expert in this field than me.

src/coordinate_system/cartesian/xzbbox.rs
XianlinSheng: Well I prefer to define all similar operators together. Imagine creating a custom dataset, you won't provide only + operator without - += -= simply because the current program hasn't used it. My intent is to provide a self-complete tool, and maybe that's also why my rust warns about unused functions but not operators - it will be used later or sooner.

src/coordinate_system/cartesian/xzbbox.rs
XianlinSheng: XZVector and XZPoint have both declared copy because they are just too simple structures, so they should behave like simple primitive types like int and float. No ownership movement is happening. If it is move-default, all XZVector in this trait should be changed to XZVector&.

But I will remove the copy trait for XZBBox next time because it will become much heavier. It will become an enum that can be rectangle or polygon shape (which is necessary for rotation - bbox needs to be rotated too, and that is not a straight rectangle anymore. Besides, how about cropping the city to a better outline shape?). When it is a polygon, relatively a lot more data will be allocated on heap to improve the performance of 'contains' method (basically a 'mask' field). All shapes will provide a circumscribed_rect method so the world_editor algorithm does not need to change a lot - looping through a larger regular area and simply skip the invalid cells. It should be acceptable to include some unused cells, at least in my numerical code the performance rarely changes.

src/coordinate_system/cartesian/xzvector.rs
XianlinSheng: It does. I separate them simply because I am used to giving class and its related methods a single file, so it will be much clearer to find the definition only related to itself. If there are particular reasons that fewer files are better, I can combine them.

src/map_editing/edit_map.rs
XianlinSheng: Sounds good. It will be changed

src/map_editing/operator.rs
XianlinSheng: If making it a trait instead, the from_json factory should become a single function operator_from_json that returns a base ptr Box<dyn Operator>, inside which it looks for operation type and further calls translate_operator_from_json. something like this?

src/map_editing/edit_map.rs
XianlinSheng: It makes sense. I can make this PR a CLI self-complete version with no extra things. I am not an expert in GUI so maybe someone else might want to decide what the window looks like and enables this feature for GUI users. I made the filename static simply because GUI users can start to play around with the feature now, even if it is not in a perfect design, and it can be improved in later PR. What's your opinion?

src/map_editing/translate/translator.rs
XianlinSheng: If we make Operator a trait instead, I'd like to keep the style consistent, so this Translator is not needed. Its method will be splitted and moved to VectorTranslator and StartEndTranslator. If keeping the Enum operator, the simplification for vector seems good to go, but I still want to make StartEndTranslator a single struct because it explicitly specifies start and end, and I can move it into this file if it is just too simple to use another.

map_editing.json
XianlinSheng: Well, I chose json simply because I find it the most common format in my experience, and many minecraft mods use json as configuration too, so maybe it is the easiest for mc users, and consistent to mc style? VS Code and Windows Terminal use json as configuration too. Another thought is that in long term I plan to develop python shell for this, so one can program a batch generation procedure or share a world by sharing the .py script, and json seems to be easy to transfer through any connections, so I made this the first format to support. I consulted ChatGPT too for json vs toml and it says that json is more common for web api, and toml for dev tools. I think in this case this config should also be one kind of api?

XianlinSheng:
> Overall AWESOME idea and implementation. I've left many notes because I'm passionate about this change and want to see it perfected and merged!
> 
> Note
> 
> I am a Rust noob. If some of these comments are wrong, feel free to tell me I'm wrong and hit resolve. No hard feelings :D
> 
> # Overall thoughts
> * I _like_ the Operator design. Very nice and flexible. How hard would it be to add things like repetition operators (e.g., repeat a transformation/rotation several times)? Looking forward to seeing how simple adding rotation will be :D
> * Please add unit tests wherever you see fit. I know they're pretty simple data structures, but it gives us additional confidence & debug info when weird edge cases arise.
> * It would also be useful to be able to pass transformations via CLI arguments. Something like `--transform translate vector 500 500` or something. That may be complex and not that great though so we can discuss separately
> 
> I'm willing to approve this* if you review my comments for important suggestions first.
> 
> * I am not the maintainer so my words have 0 weight

Repetition is very simple: repetition itself is an operator, just a special one whose operation is to call another underlying operator several times. 

Rotation on the elements is vector mathematics 101. The relatively trickier thing is the xzbbox. This thing needs to be rotated too. The idea is to make XZBBox an enum too, which can be Rect, Polygon, or even other shapes. The implementation is not difficult.

I try to add those tests following the style in your BBox PR.

Yes lets discuss the CLI implementation separately and currently just make it support reading a config file. Btw, if I decide to make the process highly programmable and able to run in batch, I choose to wrap it up as a python shell - everything connected to python will have the access to everything.

louis-e:
Woah you guys are amazing, I can't wait to see this merged!
Since there are still a few open comments, ping me when you are done from your side and ready to merge. I don't have any open comments myself, @benjamin051000 you already covered most of it with your review. Big thanks to you both!

src/coordinate_system/cartesian/xzbbox.rs
XianlinSheng: XZPoint is valid across all integers.
In principle XZBBox should be valid when both total_blocks_x and .._z > 0. ..._x is defined by point2.x - point1.x + 1. I think using assert_gt!(self.total_blocks_x(), 0) is better, same for z direction.

XianlinSheng:
- [ ] Validity check on XZBBox
- [ ] Operator trait
- [ ] Unit tests
- [ ] CLI file name argument
- [ ] File simplification (XZPoint, XZVector, and some translators if applicable)
- [ ] JSON or TOML

I reserved the places for previously merged BBox GeoCoord modules and coordinate transformations in coordinate_system. Should I cover this too or discuss it in a separate PR?

src/map_editing/edit_map.rs
benjamin051000: For now we can leave it as-is. We can make the CLI flag / GUI option in a later PR.

map_editing.json
benjamin051000: Yeah JSON is fine

benjamin051000:
As you continue work on this feel free to mark my notes as "resolved" (there's a button to do this) to clean up the PR. Thanks! Excellent work so far. It seems like you are still working on various things here so I'll wait to look at your updates

src/coordinate_system/cartesian/xzbbox.rs
benjamin051000: OK no problem

src/coordinate_system/cartesian/xzbbox.rs
benjamin051000: Ohhh. I didn't realize they were `Copy`. So they just get copied around, interesting!

Okay it seems like you've thought about this a lot, so I trust your decision.

src/coordinate_system/cartesian/xzvector.rs
benjamin051000: I would combine them to reduce duplicated code. Note: I _think_ it's possible to 
```rs
pub struct XZVector(XZPoint);
```
But I'm not sure if that will do exactly what we want.

XianlinSheng: I prefer to make them completely different, and this is to ensure the safety in math/physical algorithm using them, even if we need to duplicate and simply change names. This is a common design pattern in my area (e.g. pyansys is designed in this way) but maybe not that well-known for others, so I try to explain how it works.

In program they have exactly the same structure, but they work for different physical concepts and have different behaviours, so there are risks that user might run into algorithm errors without any warnings: point - point is still in the same structure but means displacement instead, not point anymore. Displacement - point again will be physically invalid. If you look into the operators I defined in very detail, you will notice that Point and Vector are not symmetric: 
self operations:
P - P = V; P + P ❌;
V - V = V; V + V = V;
cross operations:
P + V = P; P - V = P;
V + P ❌; V - P ❌;

So you see, Point is naturally not an addable concept, while Vector is. 2 operations only involving V comes from the closed linear space of the vector, and the other 3 are syntax sugar for widely accepted notations. 3 comes from 1 + 2, the 1 is the notation definition, and the extra 2 are derived from the V linear space. And since their relationship is quite complicated, I chose to put them in two separated files for clarity.

If making them share some names or behaviours, as the developer of this module I cannot guarantee that all operations done by the user is physically valid, e.g., what if someone wants to move an element from A to B, but made some silly mistakes without calculating the displacement but uses its original coordinate instead? (I did this once, and in my experience of helping others out from such bugs, I know this is necessary). If we do the simplification, no warnings will be thrown, but in my design the compiler will throw errors. If they need to do x += ... separately, dx has to be retrieved from the struct, and compiler will throw 'Point has no attribute dx' because dx is only a valid concept in Vector, so they know they forgot to convert coordinates into displacement. If they simply add displacement to a point, compiler will throw 'Point += Point is not implemented' because Point is not naturally addable, the addition is only valid in the context of applying a Vector displacement.

To summarize: *my intent is to ensure the safety of math algorithm based on them, and I think duplicate one file is worth it*

XianlinSheng:
A comment on BBox: I am trying to add unit test on my transformation module so I copied the example bbox coordinate in bbox test to fetch data. It failed because API did not return anything. I changed to another coordinate that has longitude > 90 and realized that this example coordinate is not in the correct order, which is instead located in the middle of Arabic Sea so of course it is returning nothing.

// Arnis, Germany (but not, this is in Arabic Sea)
assert!(BBox::new(9.927928, 54.627053, 9.937563, 54.634902).is_ok());

where

pub fn new(min_lat: f64, min_lng: f64, max_lat: f64, max_lng: f64)

I changed it to the correct order and added three more cities that have longitude > 90 or < -90, two having latitude > 0, one < 0, so that will not be confusing.

XianlinSheng:
OK, the operator module is changed to trait style as suggested; XZBBox is created using factory method and it checks for validity inside; all public member functions/data of XZBBox ensures its validity; tests are added for xzbbox, operators json reading, and translation.

XianlinSheng:
I think it's ready for review again. Most of the review comments are taken care of, except XZPoint and XZVector simplification, which I still think they should be in separate files and structures because of potential algorithm safety issue and the clarity when trying to tell apart their nuances for users.

benjamin051000 (approved):
Looks good! Thanks for submitting those changes. 

Approving. Note that I have not actually tested this code locally.

louis-e:
retrigger-benchmark

louis-e:
Hi there, first of all thanks for the great work @XianlinSheng and thanks for your extensive review @benjamin051000! I wanted to review it earlier already but got sick - so sorry for the delay. There are no more comments from my side, this is amazing to build on in the future. Is this ready to merge from your side @XianlinSheng?

src/coordinate_system/cartesian/xzbbox/xzbbox_enum.rs
```diff
@@ -0,0 +1,206 @@
+use super::rectangle::XZBBoxRect;
+use crate::coordinate_system::cartesian::{XZPoint, XZVector};
+use std::fmt;
+use std::ops::{Add, AddAssign, Sub, SubAssign};
+
+/// Bounding Box in minecraft XZ space with varied shapes.
+#[derive(Clone, Debug)]
+pub enum XZBBox {
+    Rect(XZBBoxRect),
+}
```
scd31, line 10: @XianlinSheng Hey, sorry, I know this PR is months old at this point. Just wondering, what is the purpose of this being an enum? Is there anything else that could really go in here other than a `Rect`? Just wondering as I'm adding functionality around this and want to know how generic I should keep things

XianlinSheng: Thanks for asking this question first. XZBBox is the mask of the region of generation. Normally it is a rectangle aligned with grid lines, but after rotation or other transformations this region can no longer be simply described like this. At this moment I already have a complete general polygon implementation in my branch that is necessary for rotation, so you can go to my dev-xzpoly to check out how it is designed. One can choose a form to instantiate as long as the shape can fully describe what you have now. For users of this structure, they only call the 'contains' (check a point is included). 'bounding_bbox' (circumscribed grid-aligned rectangle) method, display/add/sub traits, and possible transformations on it, so adding a new shape needs 1. 'contains' method. 2. 'bounding_bbox' method. 3. display/add/sub traits. 4. supports for your shape in all map transformation operators (it won't compile if the new shape is not covered anywhere that needs to modify the xzbbox).

The purpose of using a XZBBox is to truncate the OSM data and let it generate only inside the region. The fetched OSM data is not just the points inside the LLBBox you selected on the map but includes the complete continuous elements that are partially covered - like selecting objects in modeling software, you will select the whole object as long as the region touches it. An element may extend even much further than the llbbox dimension, and this is the reason why impl #504 found that the generation was still in the original rectangle after the rotation transform on existing coordinates - the existing data is not just the points inside the selected region, and we need to use XZBBox to restrict the generation of the redundant elements outside of it.

This design will build the necessary blocks for features like generation at non origin #29 (using rectangle), rotation #97, clip #97 (using polygon), custom shape, projection-corrected #29 (using other shapes), non-continuous space #459 (general mask). The polygon impl already has a mask field inside for the inference performance of method 'contains', and the rotation code for elevation data gives the idea to rotate a field, so a general mask box is also in fact ready-to-go, simply rearranging some existing methods.

May I know your plan so we can reach the optimal implementation?

scd31: Thanks, makes sense!

Basically I'm working on generating the world in batches, where each batch is an n chunk by n chunk square. This is to lower RAM usage, something that users have been complaining about (and even trying to script their way around batching, which kind of works, but has issues)

The way I'm doing things right now is to divide the `XZBBoxRect` up into these chunks, which are then saved to disk. Doing it this way is quite nice because it means we'll never touch the same chunk twice. Of course, that invariant doesn't hold if `XZBBox` isn't axis-aligned. Maybe you have some thoughts here?

Maybe there could be an `Intersection(XZBBox, XZBBox) variant to `XZBBox`. Then the first one could be the user-defined shape, and the second one could be the n chunk by n chunk batch.

XianlinSheng: Let me understand this situation completely and better before I give you my thoughts. Which process and which data structure is taking up the most RAM? By dividing you mean generating it in multiple steps so it won't store everything at the same time?

scd31: The goal is to batch the entire pipeline, i.e. only fetch what we need from overpass for that specific batch, then run it through the entire thing, and save it out to the region file. There will be duplicated work, especially if the batch size is too small, but it would make it feasible to create huge worlds (on the order of entire provinces/states)

XianlinSheng: OK this is a very good idea. For users with limited RAM this is the way to divide heavy jobs into smaller tasks, and for users with lots of available resources these can in reverse be done in parallel to speed things up. This is actually included in my long term plan to make it recreate large scope map, and my solution was to open a Python controlling interface and let Python to do the controlling jobs of rust pipeline. It takes all of the advantages of Python and almost avoids all of its short comes. It's easy and user friendly to control, highly programmable if want customizations, clean and simple data structure management when dealing with high-level concepts and dividing it into small rust jobs. Meanwhile, Python is slow but we only let it do controlling jobs, whose logic might be complicated enough to human but still very lightweight to Python; Python does not handle multitask well (if we want parallelization), but this architecture is to call separate rust pipelines in a program-level parallelization and Python actually works well with mpi, which I am dealing with all the time.

Python interface is better than writing a feature in rust because otherwise we are doing it in the opposite: it does not make full use of what rust is good at, and suffers from its limitations. Rust is fast, safe, but such a controlling job is fast enough even in python, and the difference is ignorable to user: even the time to fetch data will be way more than it. The limitation of rust is that it is a compiled language and has unfriendly syntax, but user should be able to manage and arrange their pipelines in their own way. If programmed in rust, the feature is strictly limited to our designed interface, either user needs to do exactly what we provide, or they have to write difficult codes and make themselves a developer, or it is just undoable.

In my field almost all industrial software has opened their controlling interface to Python. Even if they have their own scripting systems in the past, they are shifting to the same direction. This makes highly customized batch jobs possible across almost all users. Blender in CG is similar, and even Adobe software. For this world generation, I think we can do similar things, and this dividing operation actually should be one of the preset python functions. 

In details, e.g. class Job is a rust pipeline including LLBBox, transformations, elevation settings, etc., and can call the pipeline by .run(). User might want to define higher concepts like GeoRegion that is in a large enough scope but cannot be generated directly, and write a function to divide it into many Jobs, handling all settings transforms inside. What this batch feature should be such a function, but provided by us so user doesn't need to write their own and care about tricky details. 

But before deciding to do it or not, there seems to be some existing issues with world generation near the boundaries that needs to be solved first no matter what approach we choose. 1. there is nothing including the ground in the boundary chunk outside the XZBBox. 2. connectivity of roads and buildings near the boundaries needs to be guaranteed, especially when the map has some transformations like rotations.

scd31: I am pretty strongly against the idea of bringing in Python just for this. Rust has advantages over Python far beyond just speed. Its type system makes it great to work with, for instance. I personally heavily prefer Rust over Python - I would not have contributed to this project if it was written in Python, like it used to be. There's also quite a bit of complexity in having 2 languages and the advantage seems pretty dubious.

This also seems like quite a bit of scope creep for what sounds like a pretty niche feature, and will slow down the development of other features that would be more used (like the batching work I'm working on now)

cc @louis-e - it's your project so ultimately it's up to you

XianlinSheng: OK it's completely alright to have different ideas and find which is the really the optimal and suitable for this case, but that is after both proposals are really understood clearly.

The role of Python is not a second programming language, it's not part of the program, but another user interface, CLI, GUI, Python Shell, and a programmable interface has much greater abilities than these two. It will not be the developers to write Python scripts but users. Still, developers can provide some commonly needed scripts ahead of time as presets. The reason I propose this batch can be written as a Python preset is that I saw some ideas that need a programmable interface will be benefited from such a preset. Like the world generation considering projection corrections (or any approach to generate large world as you wish) in #29 needs a highly customized and programmable way to arrange the pipelines of how to combine the small jobs and form a large world. If such a preset exists, not just the users with resource problems can call the function, world designers can also call it to generate it without knowing dividing details. Projection and city design professionals will write their Python scripts and perform tests and studies to generate optimal combinations of small jobs. They don't care the rust details in generation a job, their scripts are not part of this project and they should have the freedom to choose not to focus too much on computer science but on their professionals, which is the strongest power of Python as a glue.

What you proposed makes sense and is in fact not contradictory to a Python shell. It's just a matter of on which side we realize it of the interface. I do not insist which side it should be, what I care is to make it useful to as more features as possible. So of course we can still write this in rust, but we better consider to extend the ability of the interface too (it can be just the GUI/CLI now) to make full use of this feature.

scd31: Right, I'm still concerned with the amount of lift and complexity it will add, plus how much it will slow down other feature development (we'll need to consider all sorts of edge cases as a result of allowing arbitrary user pipelines)

I definitely don't think this is something that'd be needed for #29. It needs a few extra settings at most. Maybe if they had the ability to construct arbitrary pipelines they could use them to solve #29, but it's still a lot of work, just given to the user as a Python script, instead of in the Rust core that can be more beneficial to everyone.

XianlinSheng: I understand what you mean and definitely we won't start with too big an architecture. These are just some inputs that might be useful when you decide how to expose the batch feature to user, it might be some possibilities if you are thinking how to make it work for as many cases as possible, but don't feel stressed with the information and these are not rules.

Let's back to the original question because either way some problems always need to be solved.  The current pipeline includes the map transformation and what's your current idea to make it fit in batches? The map transformation is defined for one pipeline only, if defining it to work on the original map selection, each batches will need to perform different operations, like rotating an original selection does not mean the same rotation on small jobs.

scd31: Yeah, might be something we need to figure out for the future. For now I am probably going to only support batching on axis-aligned bounding boxes

XianlinSheng: OK you can keep working on your code and I'll try to figure out a way to let this support. I'll let you know first once I have some ideas.

PR #414, opened 2025-04-20
Fix space between water blocks (meant to be 8)

Oleg4260:
Just a small change to the generation of farmlands, I noticed my mistake too late

### DIFF
--- a/src/element_processing/landuse.rs
+++ b/src/element_processing/landuse.rs
@@ -196,7 +196,7 @@ pub fn generate_landuse(
             "farmland" => {
                 // Check if the current block is not water or another undesired block
                 if !editor.check_for_block(x, ground_level, z, None, Some(&[WATER, ICE])) {
-                    if x % 8 == 0 && z % 8 == 0 {
+                    if x % 9 == 0 && z % 9 == 0 {
                         // Place water/ice in dot pattern
                         editor.set_block(
                             if args.winter { ICE } else { WATER },


src/element_processing/landuse.rs
```diff
@@ -26,7 +26,13 @@ pub fn generate_landuse(
                 GRASS_BLOCK
             }
         }
-        "farmland" => FARMLAND,
+        "farmland" => {
+            if args.winter {
+                DIRT
```
Oleg4260, line 31: Speaking of this, I intentionally did not place a SNOW_BLOCK here to make the world look a little more realistic, I'm thinking of refactoring the winter mode to remove all the SNOW_BLOCKs and put SNOW_LAYER everywhere. In my opinion, the map should generate in the way to look similar to a normal minecraft world.

benjamin051000: I agree with this. We should open a separate discussion on how to implement that, since I imagine you'd do the snow layer last after all the normal blocks are placed... 🤷🏼‍♂️

Oleg4260: Ok, we can now proceed in #415

Oleg4260:
Decided to do one more thing: I noticed that the Ground Level option is ignored when Terrain is enabled, and the MIN_Y = -62 const is used as the lowest terrain point. So I removed the const and this option is used instead, it's also -62 by default but this allows to elevate the map if needed, for example when you want to do some digging
ground_level = -62:

ground_level = -40:

ground_level = -40 with ground filling enabled:

louis-e:
LGTM, thanks!

PR #416, opened 2025-04-21
Move GUI logic to its own file

benjamin051000:
`main.rs` is an amalgamation of CLI and GUI code. Separate it out: GUI logic gets its own file since it's several functions. The CLI is one function so it can stay in main I guess.


louis-e:
Good point, the two seperate functions for CLI and GUI were already bothering me - this looks much cleaner. Code looks good to me, were you able to test the GUI? I think you mentioned once that you are only using the CLI - but I'm not sure anymore if that was you or someone else. Otherwise I'll test it real quick :)

benjamin051000:
I actually have the ability to test GUI now! I can't remember if I tested it but I will.

benjamin051000:
Okay, just ran it with 42.335390 -71.041782 42.338166 -71.035280 and it seems to generate fine!

louis-e:
Amazing, thanks a lot Benjamin! Sorry for the slight delay, I will merge this asap - not later than tomorrow. Very much appreciated🙏

PR #417, opened 2025-04-21
CI: Add cache action to preserve build artifacts.

benjamin051000:
Right now CI rebuilds from scratch for each step. It takes way longer than it needs to. This action supposedly will cache the build artifacts so they can be re-used each step. 

> [!NOTE]
> Github Copilot made this... so who knows if it actually works.

# To Do
- [ ] Consider running `cargo build --all-features` or something first, then run everything else afterwards? What's the most optimal way?


louis-e:
That's such a good idea, I didn't know Github Actions CI support caching! :)

louis-e:
I will address the --all-features in a follow up commit

Issue #384, opened 2025-02-25
[DISCUSSION] Documentation for `mcassets/`


benjamin051000:
What do these do?
Where are they from?
How are they generated?
etc.

Even just a little README in that folder with a link to get more info would be nice.

Oleg4260:
The region.template file seems to be a r.0.0.mca copied from a minecraft world, I tried creating my own flat world with a bedrock layer and replacing the file, and Arnis started crashing for some reason, so I agree, having more info would be helpful

Oleg4260:
I want to say that both level.dat and .mca files in a minecraft world can be opened with NBTExplorer, but the region.template file from the arnis project fails to open, also its content (on the left) is different from an r.0.0.mca file from a freshly created superflat world (on the right). Probably only Louis can tell us what region.template is and how its made

![Image](https://github.com/user-attachments/assets/5a33e39d-542c-4854-a3a3-d8d9c1d31d1e)

louis-e:
> When I wrote this code, only God and I understood what I did. Now only God knows.

Just kidding, it's not that bad - thanks for pointing out the missing documentation on that, I'm working on improving this at work as well as for private projects :D Isn't there a Github Wiki which can be enabled for repositories? Maybe it's a good time to start with that...

Regarding your question; here's a brief explanation of all the assets:
**icon.png**
Probably doesn't need a lot of explanation, this is the world preview image which we are inejcting into each generated world.

**level.dat**
This is a pre-made template which I generated using Minecraft 1.21.4 ([Minecraft Wiki Documentation](https://minecraft.fandom.com/wiki/Java_Edition_level_format#level.dat_format)). There aren't many customizations in there, just some basic stuff like disabled bonus chests. In the code we then modify the world name to the generated unique name, we set the LastPlayed timestamp to the current time so that the world is listed at the top in the world selection screen and we adjust the player's starting position and rotation.

**region.template** (.mca file)
This file is a bit trickier and the quote at the top of my comment might apply here. I remember it was quite a struggle to generate this file, since I was troubleshooting and trying to understand the block_states palette and data array logic at this time. As far as I remember all I did was removing all blocks for the entire region in a flat world. My goal was to get a pre-made empty world template in which block_states was initially generated by Minecraft, which is why I couldn't just use an empty world to retrieve the mca file.

louis-e:
[...]

Issue #350, opened 2025-01-28
Resources under the city


HelixeusPrim:
Hello guys! I would like to ask a question about the resources under the city. The guys and I want to survive in our city, but this is impossible due to the fact that under it there are two blocks of land and emptiness. So I would like to ask. Is it possible to somehow change this so that a city is created and the blocks under it are not deleted? Just so that a layer of earth is cut off at a certain height and then the resources of Minecraft itself go, ore, stone and bedrock below.

benjamin051000:
It's definitely possible, but I don't know if anyone has worked on it yet.

HelixeusPrim:
> It's definitely possible, but I don't know if anyone has worked on it yet. It's probably safe to assume this is on the To-do list.

Thank you very much for your answer! I'll be glad to see this in future updates

PR #421, opened 2025-04-26
Optimize ground generation + new natural elements

Oleg4260:
Some technical changes:
- Removed precalculation of ground levels for terrain mode, taking it directly from the ground object, because it didn't really improve performance and it took extra memory.
- Unified code for both terrain and flat modes by removing `if ground.elevation_enabled {...} else {...}` so the code is simpler, both modes share the same code now and ground is generated the same way in both modes, also fillground works for the flat mode, which can be useful in the future.
- Removed the spawn platform, it isn't really needed, the player always spawns on solid surface
- Always generate bedrock at MIN_Y
- Refactored the check_for_block() function, because the whitelist and blacklist worked basically the same, both returned true, and it would be hard to process whitelist for true and blacklist for false at the same time, so leaving just whitelist for simplicity

Suggestion: I think default ground_level should be changed from -62 to -61 and the region.template should be changed to generate a normal superflat world, with 1 grass, 2 dirt, 1 bedrock to match the new generation. There was a discussion in #384 about that file, but I still don't know how to change it, so @louis-e please do this if you agree with my idea

Improvements to generation:
- Added dead bush to sand regions
- Added surface types for beaches
- Implement generation of plants in grassland, scrub, heath and orchard

Todo:
- Implement tree types
- Implement better generation of wetlands

### DIFF
--- a/src/block_definitions.rs
+++ b/src/block_definitions.rs
@@ -56,7 +56,7 @@ impl Block {
             26 => "glowstone",
             27 => "granite",
             28 => "grass_block",
-            29 => "tall_grass",
+            29 => "short_grass",
             30 => "gravel",
             31 => "gray_concrete",
             32 => "gray_terracotta",
@@ -154,6 +154,8 @@ impl Block {
             133 => "ice",
             134 => "packed_ice",
             135 => "mud",
+            136 => "dead_bush",
+            137..=138 => "tall_grass",
             _ => panic!("Invalid id"),
         }
     }
@@ -277,6 +279,16 @@ impl Block {
             125 => Some(Value::Compound({
                 let mut map = HashMap::new();
                 map.insert("shape".to_string(), Value::String("south_west".to_string()));
+                map
+            })),
+            137 => Some(Value::Compound({
+                let mut map = HashMap::new();
+                map.insert("half".to_string(), Value::String("lower".to_string()));
+                map
+            })),
+            138 => Some(Value::Compound({
+                let mut map = HashMap::new();
+                map.insert("half".to_string(), Value::String("upper".to_string()));
                 map
             })),
             _ => None,
@@ -425,6 +437,9 @@ pub const DIRT_PATH: Block = Block::new(132);
 pub const ICE: Block = Block::new(133);
 pub const PACKED_ICE: Block = Block::new(134);
 pub const MUD: Block = Block::new(135);
+pub const DEAD_BUSH: Block = Block::new(136);
+pub const TALL_GRASS_BOTTOM: Block = Block::new(137);
+pub const TALL_GRASS_TOP: Block = Block::new(138);
 
 // Variations for building corners
 pub static BUILDING_CORNER_VARIATIONS: [Block; 20] = [
--- a/src/data_processing.rs
+++ b/src/data_processing.rs
@@ -143,77 +143,35 @@ pub fn generate_world(
 
     let groundlayer_block = if args.winter { SNOW_BLOCK } else { GRASS_BLOCK };
 
-    // Differentiate between terrain and non-terrain generation
-    if ground.elevation_enabled {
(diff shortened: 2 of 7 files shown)


louis-e:
Thanks for the PR! Just a few minor notes from me.

Is there any particular reason for the bedrock at MIN_Y? I guess this is a similar discussion like for the second dirt layer hahha. I just think that this will increase the generation speed unnecessarily with something that most users won't even see in this case. I'd love to hear your opinion about this.

---

> Removed precalculation of ground levels for terrain mode, taking it directly from the ground object, because it didn't really improve performance and it took extra memory.

Were you able to check the generation speed before / after with different map sizes? I don't know the exact numbers but I recall that it actually helped a lot back then. Might also be outdated by now, just making sure :)

---

> the region.template should be changed to generate a normal superflat world

I will check that!

Oleg4260:
Hello, thanks for your feedback. 

1. No, there isn't an actual important reason for that, I just think there should be a layer of indestructible block in the lowest layer to prevent holes, sometimes people like to explode the map but idk, you could move the bedrock line back to `if args.fillground {}` if you want, but I think the performance difference isn't significant. 

2. I did check generation speed before and after, after my changes it actually became a _little_ slower, but it's probably because of the bedrock layer, also before my changes it would take some time to process the ground heights before actual generation started, I guess it's the same amount of time now, but as a part of the ground generation time. Also the main reason I did this is because I wanted the ground to be generated by the same code in both states of args.terrain.

Oleg4260:
After getting some sleep I found out that there's still a way to pre-calculate ground levels, checking right now if it's faster than without it
image

Oleg4260:
Here are the results:

Bbox: 49.383714 15.605736 49.430403 15.689163 (31.3 km²)

With precalculation:
    Terrain on: 1:24
    Terrain off: 0:14

No precalculation:
    Terrain on: 1:28
    Terrain off: 0:13

As we can see, it's nearly the same, so that's why I think we don't really have to pre-calculate ground levels

Oleg4260:
I am going to add some improvements to generation of some natural surfaces, so drafting this temporarily

Oleg4260:
Added generation of dead bush in natural:sand, this is how some [random desert in Libya](https://maps.app.goo.gl/zrQ8F5UjEm3fBYkk6) looks like now.
I am currently working on generation of some nice grasslands, scrubs, woods, then I will implement generation of wetlands, like marshes, bogs etc.
I don't have much of free time now, so I will continue later

Oleg4260:
Grassland, scrub and wood somewhere near Izium, Donestk oblast. Not final result, still needs some adjustment.

Oleg4260:
Final version of `natural:scrub` is below. Didn't want to make it more complicated, since this isn't a very significant element

Oleg4260:
An orchard.

louis-e:
Very cool, thanks a lot! I'm always happy about visual enhancements :)
Just so you know, I might change the bedrock layer again in the future - maybe by aligning it directly with the elevation level instead of at a fixed Y coordinate.

Oleg4260:
> maybe by aligning it directly with the elevation level instead of at a fixed Y coordinate.

If you mean that bedrock might be higher than -64 in the future, I don't think it's a good idea. I am trying to make the generated map be as close as possible to a normal minecraft world. while also keeping it realistic and corresponding to the OSM data. There are suggestions like for example #350, and I really like the idea of making it possible to survive on a generated map, so I'm slowly moving towards adding an option of generating resources, and that's why everything between the ground level and -64 has to be filled with stone, only with bedrock being at the lowest level

Oleg4260:
@louis-e just a ping to be sure that you are aware of my reply

louis-e:
Alright, then let's keep it as it is for now - I also really need to get some sleep now haha!
Also I'm really looking forward to the underground structures. I really want to have them included as well at some point!

Issue #415, opened 2025-04-21
[DISCUSSION] Reworking the winter mode


Oleg4260:
I've been thinking about this for a while, because I don't like how the winter mode works currently.
The goal is:
- Grass blocks should not be replaced with snow blocks, but should be covered with snow layer above them, just like most objects.
- Roads should not be covered with snow, maybe except some minor roads without solid surface
- Water should be either replaced with ice or removed (in artificial objects like fountains)
- Minor plants like grass and flowers should not be generated, but replaced with snow layer instead
- The code has to be clean, fast, reliable, and compatible with adding new features

The first small step for this was already done by me in #403, now in winter mode there are dirt blocks covered by snow layer blocks on farmlands, water in farmlands is frozen, and most lakes are frozen.

This discussion was created to talk about the best way to implement this.

Oleg4260:
Currently I see two ways to implement generation of snow layer on the ground:
- In the data_processing.rs, by detecting the highest block and placing snow layer above it, similarly to how dirt is generated under grass
- In every file of element_processing

Both options, especially the first one, might have a problem of roads being covered, we need to choose one of these, then invent a way to prevent roads and some other objects from being covered in snow

benjamin051000:
I mean, roads _do_ get covered with snow in winter months... but It would probably not be ideal for this program. 

The snow step could take in coords of road blocks and just skip them? I'm not sure how roads are represented internally but I imagine it boils down to a list of coordinates

benjamin051000:
Note that in winter biomes, the water will automatically freeze after some amount of time. I guess snow will fall naturally too which will create snow layers? But it can take a while I guess

Oleg4260:
Right, snow falls on roads like on any other surface, but the local authorities usually keep them clean to prevent issues with traffic, so it's the most realistic variant. I'm not sure about biomes, AFAIK the program cannot change the biome directly in the world, also I don't know how they work in Minecraft and if it's possible to change them, maybe we could use a different region.template file of a world with winter biome, but in that case we need to know more about that file, and your discussion #384 still didn't get any answers

louis-e:
> I mean, roads do get covered with snow in winter months... but It would probably not be ideal for this program.

> Right, snow falls on roads like on any other surface, but the local authorities usually keep them clean to prevent issues with traffic, so clean roads in winter is the most realistic variant, also it looks much better than snow-covered roads.

I agree with this, roads shouldn't be covered with snow.

[...]

PR #422, opened 2025-04-26
Remove winter mode as discussed in issues/415

louis-e:
- Removed winter / snow mode as discussed in https://github.com/louis-e/arnis/issues/415 since it only adds complexity in future

### DIFF
--- a/src/element_processing/landuse.rs
+++ b/src/element_processing/landuse.rs
@@ -19,27 +19,9 @@ pub fn generate_landuse(
     let landuse_tag: &String = element.tags.get("landuse").unwrap_or(&binding);
 
     let block_type = match landuse_tag.as_str() {
-        "greenfield" | "meadow" | "grass" => {
-            if args.winter {
-                SNOW_BLOCK
-            } else {
-                GRASS_BLOCK
-            }
-        }
-        "farmland" => {
-            if args.winter {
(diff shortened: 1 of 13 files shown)


src/element_processing/landuse.rs
```diff
@@ -204,16 +176,7 @@ pub fn generate_landuse(
...
-                            if args.winter { ICE } else { WATER },
-                            x,
-                            ground_level,
-                            z,
-                            Some(&[FARMLAND, DIRT]),
-                            None,
-                        );
-                    } else if args.winter {
-                        editor.set_block(SNOW_LAYER, x, ground_level + 1, z, None, None);
+                        editor.set_block(WATER, x, ground_level, z, Some(&[FARMLAND, DIRT]), None);
```
Oleg4260: Should be just FARMLAND here, there won't be any DIRT without the winter mode

Oleg4260 (changes requested):

src/element_processing/landuse.rs
louis-e: Good catch, thanks for the fast review! :D

louis-e:
Merging this for now, I'll look out for any objectives in https://github.com/louis-e/arnis/issues/415 and revert the removal if wanted.

PR #423, opened 2025-04-27
Simplify elevation retrieval in each element processor

louis-e:
Instead of retrieving the elevation individually in each element processor, we now do it directly in world_editor.rs. Work in progress, not done yet (especially for buildings.rs and amenities.rs)


louis-e:
Before
image

After
image

Oleg4260:
So the y argument is relative against ground level now? Interesting solution, not sure if I like it, but let's see if it's faster

Oleg4260:
I think this might cause problems where you want to place a block on absolute height, not relative to the ground level, especially ground filling and bedrock layer

louis-e:
> So the y argument is relative against ground level now? Interesting solution, not sure if I like it, but let's see if it's faster

My main intention with this PR is to simplify the element processors. Would be even better if it turns out to make a significant difference in speed.

> I think this might cause problems where you want to place a block on absolute height, not relative to the ground level, especially ground filling and bedrock layer

Good point, since absolute block setting is used rarely - does a seperate function for that sound okay for you? I'll commit a draft of that real quick

Oleg4260:
> Good point, since absolute block setting is used rarely - does a seperate function for that sound okay for you? I'll commit a draft of that real quick

Yeah, now it's good

Oleg4260:
There's one thing that makes me worry. Your PR changes a lot in the code, and if you merge it first then I would have to solve merge conflict. Will you merge my PR #421 first or yours #423 first?

louis-e:
I'll merge yours first and then follow up with this PR! :)

Oleg4260:
@louis-e Hello again, looks like we abandoned this discussion. And I noticed another bug caused by this PR: buildings on hills now look goofy because their height is relative. Also you already made a poc for basements under buildings (discussed in #406), will you push it into main or can you at least just fix this issue, please?

Oleg4260:
P.S. Don't forget that there are buildings with min_level tag, they don't need a basement, it's already implemented, just don't break it :)
image

louis-e:
Thanks for the comments and then ping @Oleg4260! Sometimes I don't have the overview about all updates in here since my email inbox gets spammed with tons of mails per day hahha!

> min_level tag

I noticed a bug with that tag as well as with building bridges since the merge of this elevation retrieval refactoring, I will take care of that!

> buildings on hills now look goofy because their height is relative.

I actually liked this approach a lot. I want to get some more impressions of it, also in comparison to the stone foundation approach. Maybe we can do a vote on which one people like more with some example images in the next few days!

Oleg4260:
> I actually liked this approach a lot. I want to get some more impressions of it, also in comparison to the stone foundation approach. Maybe we can do a vote on which one people like more with some example images in the next few days!

I mean, buildings in real life are never built like this, they always have a correct geometric shape

Oleg4260:
> stone foundation approach

Also as I said in the discussion, I think it's best to make it of the same block as wall block instead of cobblestone

Issue #418, opened 2025-04-23
[FEATURE] Add subways and tunnels


GMart:
**Idea**
Right now the code skips tunnels and subways.
I don't know Rust very well. I think this can be added:
If the tag includes tunnel or subway, and terrain is turned on and fill ground is turned on, then
respect the layer tag, and render the rails underground. Include one block of air above the rail.
Any rail stations can be rendered as grey terracotta or concrete next to the rail line.

**Additional context**

GMart:
Here is what I have so far. Rails generate underground, but there is discontinuities where the `level` tag changes. I think if we add smoothing to the y axis, it should fix it.
Both screenshots taken from NYC, Lower Manhattan with terrain and fill ground on.
![Image](https://github.com/user-attachments/assets/799545bc-27b9-4701-add1-80d2a2d53a04)
Second showing underground station
![Image](https://github.com/user-attachments/assets/7bd89c55-0aa3-4be8-b0cb-71d6cef5501f)

Let me know if this is a good start to a PR. it's current in my fork: [Here](https://github.com/GMart/arnis/tree/add-subways-and-layers)

GMart:
Roads also generate underground now, but right now it's only 1 block tall.

![Image](https://github.com/user-attachments/assets/6d5626b2-23f5-49b6-bc61-c6a2f9e7751a)

GMart:
Sleepers now generate underground, with the gravel not, because there is usually just the stone between sleepers.
Underground roads and paths now generate correctly.
This screenshot shows a tunnel between two sides of a subway platform. Not perfect, but the layers show correctly.

![Image](https://github.com/user-attachments/assets/ec56e452-4f48-4396-8f69-3b32d7a4f25f)

PR #425, opened 2025-04-27
Update leisure and highway element processing

GMart:
Hi Louis,
This is my first time contributing to a Rust project. 
I noticed a few things I could add while still learning the Rust syntax.

- Fix 'Mincraft' typo (Code still works the same so not sure what that prefix does) 
- Add additional types for leisure
    -  I tested this on an ice rink, but it still generated the normal building over it. Ice-rink generation may need to be moved to buildings.rs   
    - Will test slipway and nature reserve
- Add additional types for highway

### DIFF
--- a/src/element_processing/highways.rs
+++ b/src/element_processing/highways.rs
@@ -131,6 +131,22 @@ pub fn generate_highways(editor: &mut WorldEditor, element: &ProcessedElement, a
                     block_type = GRAY_CONCRETE;
                     block_range = 2;
                 }
+                "secondary_link" | "tertiary_link" => {
+                    block_type = BLACK_CONCRETE;
+                    block_range = 1;
+                }
+                "escape" => {
+                    block_type = SAND;
+                    block_range = 1;
+                }
+                "steps" => {
+                    block_type = OAK_SLAB;
+                    block_range = 1;
+                }
+                "ladder" => {
+                    block_type = LADDER;
+                    block_range = 1;
+                }
                 _ => {
                     if let Some(lanes) = element.tags().get("lanes") {
                         if lanes == "2" {
--- a/src/element_processing/highways.rs
+++ b/src/element_processing/highways.rs
@@ -143,10 +143,7 @@ pub fn generate_highways(editor: &mut WorldEditor, element: &ProcessedElement, a
                     block_type = OAK_SLAB;
                     block_range = 1;
                 }
-                "ladder" => {
-                    block_type = LADDER;
-                    block_range = 1;
-                }
+
                 _ => {
                     if let Some(lanes) = element.tags().get("lanes") {
                         if lanes == "2" {
--- a/src/element_processing/highways.rs
+++ b/src/element_processing/highways.rs
@@ -132,14 +132,17 @@ pub fn generate_highways(editor: &mut WorldEditor, element: &ProcessedElement, a
                     block_range = 2;
                 }
                 "secondary_link" | "tertiary_link" => {
+                    //Exit ramps, sliproads
                     block_type = BLACK_CONCRETE;
                     block_range = 1;
                 }
                 "escape" => {
+                    // Sand trap for vehicles on mountainous roads
                     block_type = SAND;
                     block_range = 1;
                 }
                 "steps" => {
+                    //TODO: Add correct stairs respecting height, step_count, etc.
                     block_type = GRAY_CONCRETE;
                     block_range = 1;
                 }
--- a/src/element_processing/highways.rs
+++ b/src/element_processing/highways.rs
@@ -140,7 +140,7 @@ pub fn generate_highways(editor: &mut WorldEditor, element: &ProcessedElement, a
                     block_range = 1;
                 }
                 "steps" => {
-                    block_type = OAK_SLAB;
+                    block_type = GRAY_CONCRETE;
                     block_range = 1;
                 }
 
--- a/src/element_processing/leisure.rs
+++ b/src/element_processing/leisure.rs
@@ -15,8 +15,9 @@ pub fn generate_leisure(editor: &mut WorldEditor, element: &ProcessedWay, args: 
 
         // Determine block type based on leisure type
         let block_type: Block = match leisure_type.as_str() {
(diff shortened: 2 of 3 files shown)


XianlinSheng:
Seems that you need to run cargo fmt to format the files, and also run cargo clippy to check for any verbose or discouraged logic.

src/element_processing/leisure.rs
```diff
@@ -15,8 +15,8 @@ pub fn generate_leisure(editor: &mut WorldEditor, element: &ProcessedWay, args:
 
         // Determine block type based on leisure type
         let block_type: Block = match leisure_type.as_str() {
-            "park" => GRASS_BLOCK,
-            "playground" | "recreation_ground" | "pitch" => {
+            "park" | "nature_preserve" | "garden" => GRASS_BLOCK,
```
Oleg4260: It's "nature_reserve", not preserve

Oleg4260: Also you probably want to add that tag to the line 84, to make trees generate

Oleg4260 (approved):

src/element_processing/leisure.rs
GMart: Thanks, fixed!

src/element_processing/leisure.rs
```diff
@@ -15,21 +15,27 @@ pub fn generate_leisure(editor: &mut WorldEditor, element: &ProcessedWay, args:
...
-            "playground" | "recreation_ground" | "pitch" => {
+            "park" | "nature_reserve" | "garden" | "disc_golf_course" | "golf_course"
+            | "soccer_golf" => GRASS_BLOCK,
+            "playground" | "recreation_ground" | "pitch" | "beach_resort" | "dog_park" => {
                 if let Some(surface) = element.tags.get("surface") {
                     match surface.as_str() {
                         "clay" => TERRACOTTA,
                         "sand" => SAND,
                         "tartan" => RED_TERRACOTTA,
+                        "grass" => GRASS,
```
Oleg4260: Should be GRASS_BLOCK, because GRASS stands for short_grass, which is not a solid block. 

Also I saw that you invited me to your fork, want help with something? I don't know where is it better to contact you so writing here

src/element_processing/highways.rs
```diff
@@ -131,6 +131,22 @@ pub fn generate_highways(editor: &mut WorldEditor, element: &ProcessedElement, a
...
+                "escape" => {
+                    block_type = SAND;
+                    block_range = 1;
+                }
+                "steps" => {
+                    block_type = OAK_SLAB;
+                    block_range = 1;
+                }
+                "ladder" => {
+                    block_type = LADDER;
```
Oleg4260: This might be a problem, have you tested how it generates? If it's bad, better remove it

Oleg4260 (changes requested):

src/element_processing/highways.rs
GMart: Yup, looks like it's only used on 422 nodes and 600 ways worldwide. I will remove for now.

src/element_processing/leisure.rs
GMart: OK, I will fix the GRASS_BLOCK. 
If you want to see what I have done with underground structures, feel free, otherwise you can ignore. I might need help with some things. Should I post them in the issues here, or in my fork?

Oleg4260: Better here, as a discussion in issues or a draft PR, depending on what your problem is, so more people can see it and give an advice

louis-e:
Hi there, thanks a lot for your first contribution! :)
I will test everything locally and then follow up with questions or merge it directly.

> Fix 'Mincraft' typo

Woah, I have no clue when that typo was introduced. Never noticed that before in the code. Thanks!

> what I have done with underground structures

Nice, underground structures? If you want to get this PR merged sooner, feel free to open that in a seperate PR. Otherwise you can also mark this PR as a draft and push your other updates into this.

GMart:
@louis-e, 
Thanks! 
The underground structures are in another one of my forks, here: https://github.com/GMart/arnis/tree/add-subways-and-layers and I have some screenshots here: #418 
I tried to separate the PRs ahead of time.

src/element_processing/highways.rs
```diff
@@ -131,6 +131,19 @@ pub fn generate_highways(editor: &mut WorldEditor, element: &ProcessedElement, a
...
                 }
+                "secondary_link" | "tertiary_link" => {
+                    block_type = BLACK_CONCRETE;
+                    block_range = 1;
+                }
+                "escape" => {
+                    block_type = SAND;
+                    block_range = 1;
+                }
+                "steps" => {
```
louis-e, line 144: https://wiki.openstreetmap.org/wiki/Tag:highway%3Dsteps
Steps need to be implemented in a different way. With the current implementation, they would all be generated on ground level and would not represent an elevation. I couldn't find any in my world, but here is an example how this will probably look like.
image


So my proposal is to remove lines 142-146 and take care of implementing steps someday in the future. As soon as this is resolved I can merge it!

GMart: Yes, I agree. I just wanted something different from the default road texture. Can I put oak planks for now?
Looks like we can use step_count (/2)or level to determine the height of the stairs, incline=up or incline=down to determine the direction, and width to determine the width of the stairs. We can also use surface= to determine the surface type. 
Speaking of, it looks like surface= and material= are very commonly used tags, maybe we can get a standard helper function for determining the blocks used for those tags. Right now it's a lot of repeated code to determine those blocks for each type of element.

louis-e: Feel free to implement that and provide me some images if you want! I just think that there might be a few edge cases which need to be testeed extensively. Otherwise let me know if you just want to remove it for now. Thanks for the input! :)

GMart: Ok, I have set to GRAY_CONCRETE for now. Left it in to remind me to update it later. I'll make a new PR for adding stairs.

GMart:
retrigger-benchmark

louis-e:
FYI the trigger updates the last benchmark comment.
LGTM, thanks! :)

PR #426, opened 2025-04-28
Fixes for pull/423

Oleg4260:
- Fix custom layers being generated over ground level
- Remove relative_to_absolute_y() because there's already get_absolute_y() that serves the same purpose
- Remove unnecessary dirt filling, remove duplication of bedrock line
- Fix crash caused by unicode symbols in place names

### DIFF
--- a/src/data_processing.rs
+++ b/src/data_processing.rs
@@ -156,9 +156,6 @@ pub fn generate_world(
 
             // Fill underground with stone
             if args.fillground {
-                // Set bedrock at the absolute Y = MIN_Y
-                editor.set_block_absolute(BEDROCK, x, MIN_Y, z, None, Some(&[BEDROCK]));
-
                 // Fill from bedrock+1 to 3 blocks below ground with stone
                 editor.fill_blocks_absolute(
                     STONE,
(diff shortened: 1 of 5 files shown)


Oleg4260:
Hi @louis-e , just noticed a few mistakes you made in your last PR, for example you forgot to put - in the lines that generate custom underground layers, which caused them to generate over ground level, and decided to remove a function for getting absolute y that simply calls another function, so it's basically useless. Also the program started crashing when I tried to generate my hometown, because its name contains the ї symbol, which is a part of the Unicode, but now it's fixed

Oleg4260:
Btw, remember this [performance test](https://github.com/louis-e/arnis/pull/421#issuecomment-2833410709)? It now takes only 18 seconds to generate, which is over 4 times faster!

louis-e:
Thanks for the extensive testing, I'm glad you caught some mistakes - let me know if you notice more! :)

> Btw, remember this https://github.com/louis-e/arnis/pull/421#issuecomment-2833410709? It now takes only 18 seconds to generate this bbox with terrain enabled, which is over 4 times faster! You really improved the speed of ground generation amazingly

Woah nice, I didn't expect it to make such a big difference!

PR #442, opened 2025-05-05
Add subways and layers

GMart:
I see there is already a PR adding subways. I think I can merge that one in, as the look is a little better than mine. 
I have also added subway stations and train stops.

### DIFF
--- a/src/data_processing.rs
+++ b/src/data_processing.rs
@@ -40,8 +40,74 @@ pub fn generate_world(
 
     let progress_increment_prcs: f64 = 49.0 / elements_count as f64;
     let mut current_progress_prcs: f64 = 11.0;
-    let mut last_emitted_progress: f64 = current_progress_prcs;
+    let  _last_emitted_progress: f64 = current_progress_prcs;
 
+    process_pb.finish();
+
+    // Generate ground layer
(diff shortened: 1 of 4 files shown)


GMart:
Screenshot here: https://github.com/louis-e/arnis/issues/418#issuecomment-2844832052

GMart:
Current bugs with this fork: 

-  The progress bar is messed up since I switched the order of generation, ground generation happens first.

- Underground roads can intersect with subways but the subway walls are always generated in the way. It may not be possible to determine when the path should be open to the subway. Usually, stations generate correctly, next to the subway line.
- Subways can look weird with both terrain and fill ground on.
- Subway entrances don't always generate on the surface.

GMart:
When I have time, I will separate this out into two PRs, one for the Subways and underground, and one for modifying how the generation works so that underground structures generate with fill ground on.

louis-e:
Hi GMart, thanks a lot for your work!
FYI I'm currently working on finishing up everything for the next major release. Originally I was planning on including the subway feature for this release, but I also struggled with integrating it, so I appreciate you spending time on that. Depending on when the release will happen, we can include this either in the major release or as a minor release. :)

PR #444, opened 2025-05-10
Fix scaling of roads and building heights

### LINKED ISSUE #443: [BUG] Buildings, roads don't scale downwards
**Describe the bug**
When running arnis with a scale <1 (e.g., 0.1 for 1/10th the size), roads don't appear to scale downwards. Also, tall buildings seem to render incorrectly, with artifacts of the building floating in the air.

Command:
```
❯ cargo run --release --no-default-features -- --bbox "40.698861,-74.021416,40.831996,-73.919792" --terrain --scale 0.1 --path "/home/benjamin/.var/app/org.prismlauncher.PrismLauncher/data/PrismLauncher/instances/Fabulously Optimized/minecraft/saves/New World/"
```

**Screenshots**

Image
See how the roads are super wide? They should be 1/10th the size, so probably only a block wide.

Image
Tall buildings (this is the financial district in Mahnattan, New York, so tons of skyscrapers) have bits floating in the air. So something weird is happening there, too.

GMart:
Fix road widths to respect scaling. Buildings now respect scale for start_y_offset, fixing most floating buildings
Fix #443 

TODO:

- [x] Fix Road stripes, they are showing too much white on the roads from stripes, since stripes aren't scaled yet

Update: Fixed stripes. In a future PR we need to look at other things that have height when world scale is < 1, like trees


GMart:
retrigger-benchmark

louis-e:
Perfect, thanks a lot! :)

PR #446, opened 2025-05-15
Small data retrieval improvements

akx:
Just some papercuts I noticed while trying out the CLI:

* `args.run()` is an unfortunate place to do argument validation; can just as well do it while parsing.
* I messed up my bbox the first time around and got an unhelpful error. No more!
* It's useful to be able to always choose to save the Overpass data to a file, instead of only when happening to use `--debug`, so that's now a thing (`--save-json-file`).


louis-e:
Great improvements, thanks a lot for your contribution! :)

PR #447, opened 2025-05-15
Small CI fixes

akx:
* use https://github.com/Swatinem/rust-cache for caching
* don't run formatting checks & clippy thrice

(See https://github.com/akx/arnis/pull/1 for this running in my fork)

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -47,16 +47,7 @@ jobs:
         sudo apt install -y libgtk-3-dev build-essential pkg-config libglib2.0-dev libsoup-3.0-dev libwebkit2gtk-4.1-dev
         echo "PKG_CONFIG_PATH=/usr/lib/x86_64-linux-gnu/pkgconfig" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


benjamin051000:
We already have the cargo cache set up. Why change to something else?

akx:
> We already have the cargo cache set up. Why change to something else?

Swatinem/rust-cache is simpler (in the CI definition file) and basically just works. That bit could be reverted if needed, though.

.github/workflows/ci-build.yml
```diff
@@ -1,30 +1,21 @@
 name: CI Build
 
-# Trigger CI on pull requests and pushes to main, when relevant files change
+# Trigger CI on pull requests when relevant files change, and pushes to main
 on:
   pull_request:
-    branches:
-      - main
     paths:
+      - '.github/**'
```
louis-e, line 7: Is there any specific reason to trigger the CI if something in .github/ changes?

akx: Yes - so changes to the CI configuration (which live in `.github`) get tested in PRs that change the configuration. 😄

PR #450, opened 2025-05-16
Data parsing fixes

akx:
* Makes `parse_osm_data` take ownership of the raw data instead of the caller cloning it. IOW, the parser function transmutes the raw data to the parsed stuff.
* Similarly, `SplitOsmData` now does the work of splitting the elements by type, so the loops that handle them don't need to check for types themselves (and also can work without references, on the data itself.) (Future work could improve things there for less cloning.)
* Makes writing the debug file use a buffered writer, which makes it much, much faster.


louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

louis-e:
LGTM, thanks a lot! :)

PR #451, opened 2025-05-16
Some overflow fixes

akx:
I tried Arnis with a large area that was painfully slow on `--debug`, and would just hang on `--release`.

I mustered enough patience to wait for `--debug` to do its thing, only to see it panic on a overflow.

This PR fixes two of the overflow panics I found, and also turns on `overflow-checks` in release builds so they won't hang in similar situations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ readme = "README.md"
 
 [profile.release]
 lto = "thin"
+overflow-checks = true
 
 [features]
 default = ["gui"]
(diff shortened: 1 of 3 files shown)


benjamin051000:
What performance penalty does overflow checking have in release?

akx:
> What performance penalty does overflow checking have in release?

Presumably some, since it's not enabled by default in `release` 😅 I didn't measure.

GMart:
retrigger-benchmark

akx:
Btw, I wouldn't necessarily trust the benchmark results - [AISI, it](https://github.com/louis-e/arnis/blob/main/.github/workflows/pr-benchmark.yml)'s measuring how long the implicit `cargo build` for a `cargo run` takes?

It'd be better to `cargo build --release --no-default-features` first, and then benchmark a couple times with e.g. [hyperfine](https://github.com/sharkdp/hyperfine) to make things more sound statistically.

I can PR that if there's interest.

louis-e:
retrigger-benchmark

PR #456, opened 2025-05-18
Switch map provider from Mapbox to OSM

louis-e:


akx:
FWIW, I've had great success with https://openfreemap.org/ 's vector tiles :)

PR #457, opened 2025-05-19
CI: fix benchmark time to not include build

akx:
Timing `cargo run` means also timing the build if one doesn't exist.

This splits the benchmark CI to build first, then just time the compiled binary.

A follow-up could be to use https://github.com/sharkdp/hyperfine so the benchmark is run a couple of times, but then you'd need to take into account how many times Hyperfine decides to run the benchmark, which is not something I want to implement in Bash math.

I adjusted the baseline duration accordingly in this PR.

### DIFF
--- a/.github/workflows/pr-benchmark.yml
+++ b/.github/workflows/pr-benchmark.yml
@@ -22,7 +22,6 @@ jobs:
         uses: dtolnay/rust-toolchain@v1
         with:
           toolchain: stable
-          components: clippy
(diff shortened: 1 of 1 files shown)


benjamin051000:
Is the hardware guaranteed to be the same across benchmark runs? I have no idea how GH actions runners are chosen/provisioned.

louis-e:
retrigger-benchmark

PR #462, opened 2025-05-24
Add new desktop icon

louis-e:
I would like to collect some feedback on this desktop icon. Should we change it? Should we keep it?
image


louis-e:
Good idea, thanks - I will experiment with that a bit! Yes it is AI generated, since I don't have the skills to do that on my own.

PR #465, opened 2025-05-29
Buildings Overhaul

louis-e:
WIP

- Generate buildings with foundation instead of aligning with the elevated ground
- Implement roofs (https://wiki.openstreetmap.org/wiki/Key:roof:shape)
- Window variations
- Interior generation
- Refactored building color block mapping

image


louis-e:
retrigger-benchmark

src/element_processing/buildings_interior.rs
```diff
@@ -0,0 +1,276 @@
...
+        for z in interior_min_z..=interior_max_z {
+            for x in interior_min_x..=interior_max_x {
+                // Skip if outside the building's floor area
+                if !floor_area_set.contains(&(x, z)) {
+                    continue;
+                }
+
+                // Map the world coordinates to pattern coordinates using modulo
+                // This creates a seamless tiling effect across the entire building
+                let pattern_x = ((x - interior_min_x) % pattern_width + pattern_width) % pattern_width;
```
GMart: Idea: Modify these two lines to be like: ((z - interior_min_x) % pattern_height + pattern_height + floor_index as i32) % pattern_height 
This adds a shift every additional floor so looking at the side isn't the same every level except the ground floor.

GMart: Would look like this when viewed from the side:

GMart:
Here is Notre Dame Cathedral in Paris. it almost works!

GMart:
However this is across the island, roof of the Conciergerie: 

You should be able to use the roof:height and roof:orientation attribute instead of calculating it. https://www.openstreetmap.org/way/248480552

louis-e:
Thanks for your reviews @GMart, will respond to them asap! :) Just to let you know the PR isn't finished yet so a lot of stuff still looks weird haha! But I'm glad you like it so far!! :)

src/element_processing/buildings_interior.rs
louis-e: Great point, thanks a lot!

louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

PR #466, opened 2025-05-30
Update barrier generation

GMart:
Fix typo with building floor variations color map, and add oak trapdoor block
Update barrier generation: Separate out ways and nodes. Add gates and stiles, etc as nodes.
Update metal and related barriers to IRON_BARS


GMart:
Gates currently show up like this with hedges and other barriers, not sure why. I think the barrier way is processed after the node. 

There should be air above the trapdoor.

louis-e:
Sorry for the delay, I was on a business trip and got sick afterwards - looks good to me, will merge the PR now! :)

PR #474, opened 2025-06-10
Rearrange geocoord and bbox into coordinate_system in the same style as cartesian point and xzbbox

XianlinSheng:
1. bbox::BBox -> coordinate_system::geographical::LLBBox
2. geocoord::GeoCoord -> coordinate_system::geographical::LLPoint
Same way as cartesians are defined
4. lat_lon_to... -> coordinate_system::transformation::CoordTransformer.transform_point(llpoint)
This stores the data inside, no need to pass lots of parameters when calling function. Unit tests have been done comparing this transform call and legacy function.

New spawn coordinate code:
1. Change the coordinate tranformation to the transformer.transform_point(...) style
2. Change the point valid check to clearer llbbox.contains(...) style
3. *** Is the code 'update_player_position' called in gui? my cargo says it is unused and spawn location seems to be not working using what compiled by cargo run. If something wrong is in my env I need to correct it.
4. Spawn location should support corresponding change after applying map_transformation, and this could be done by modifying this coordinate transformer.


louis-e:
Hi there! Sorry for the delay, I was on a business trip and got sick afterwards - so I had to catch up with a few PRs. I'll review this next after the merge conflicts are resolved. Thanks a lot for your effort, very much appreciated! :)

louis-e:
retrigger-benchmark

XianlinSheng:
Take your time, I'm not in hurry. Let me know your comments when you start working on this.

louis-e:
Hi, first review round! :)

>  Is the code 'update_player_position' called in gui?

Yes, the user can select the spawn position via the location marker in the GUI. So we'd need to keep that in.
image
I've seen the issue that cargo reports it as an unused function and I actually don't know why, since it is used in the codebase.
image
_(Main branch)_

> spawn location seems to be not working using what compiled by cargo run.

This is probably because you didn't select a spawn location then.

louis-e:
retrigger-benchmark

louis-e:
I added update_player_position again and resolved the merge conflicts. Looks good to me now, thanks a lot for your work! :)

PR #494, opened 2025-07-21
Better waterway parsing

louis-e:


louis-e:
retrigger-benchmark

louis-e:
Still having some issues with differentiating waterway elements. Example below (LineString vs closed polygon)
image

image

PR #495, opened 2025-07-23
Add man_made tagged elements

louis-e:


louis-e:
retrigger-benchmark

louis-e:
Benchmark time increased recently due to new added elements. I will investigate and adapt the reference time soon

PR #523, opened 2025-08-23
fix typo in args

scd31:
Took me a while to figure out why the API wasn't returning any data  (:


louis-e:
Good catch, thanks! :)

PR #532, opened 2025-08-24
Update README.md

sralmai:
Fix lat and long ordering in the command line build example.


louis-e:
Good catch, thanks! :)

PR #533, opened 2025-08-24
Nix development fix/improvements

scd31:
- Fixed the nix build
- Moved into root directory where imo it belongs
- Added .envrc


louis-e:
Thanks for the PR, a few questions since I never used Nix :D
1) Is it worth documenting the Nix setup and direnv usage in the Wiki, or do most Nix users know how to deal with that?
2) Since this adds a few more files in the root directory, thus making the README to non-Github users less obvious (they have to scroll down even more to find the README with valuable information on the Github page), is there a way to reduce the amount of files in the root directory?

I have no clue how many people are actually using Nix, but I'm totally not against adding it. Just want to avoid losing interested people due to not finding the valuable information in the README. Thanks a lot scd!! :)

scd31:
1. Personally I don't think so. For Nix people it's just a matter of doing `nix develop` if they don't use .envrc, or simply changing into the directory if they do. 
2. Just curious, do you have a target audience or platform in mind? As a developer the `README.md` file is the first one I look for when in a new repository. If we do really want to reduce the number of top-level files, it seems like `example_transformations.json`, `taginfo.json`, and `tauri.conf.json` could all be moved (to be clear, I don't think we need to do this)

louis-e:
1) Alright sounds good :)
2) I already prepared a PR which cleans up the root directory a bit, I'll merge it after all other PRs to avoid merge conflicts on their ends.
_"Just curious, do you have a target audience or platform in mind?"_: From what I've seen, most people using Arnis have no development knowledge at all. There are even some fake websites which look official, a lot of people actually accidentally link those instead of this Github site. So I don't want to make it worse by hiding the README even more. However it should be fine with my clean up PR, so I'm just gonna merge this! :)

PR #534, opened 2025-08-24
fix dynmap, and maybe other things

scd31:
Dynmap expects `data` to not exist if it would be of length 0. Other software may expect the same so I fixed it here rather than in Dynmap

I also deduped some code while I was at it


louis-e:
retrigger-benchmark

louis-e:
Thanks, I will look into this soon - I'm currently on vacation! :)

louis-e:
Thanks for the PR, I think this is very useful since a few people already complained about this issue. I just added the block property insertion again since that was accidentally removed. Ready for merge

Issue #537, opened 2025-08-25
Arnis crash at fetching elevation stage


daninskip:
Arnis crashes out when it gets to fetching elevation stage. It doesnt matter what area or what size area I choose, it happens. I managed to get it to work the first time I used the new version (2.3.0), but then it has happened every time. It runs fine if I turn off the terrain option.

latest log report:
[2025-08-25][22:57:31][arnis::gui][ERROR] Application panicked: PanicHookInfo { payload: Any { .. }, location: Location { file: "src\\ground.rs", line: 27, column: 14 }, can_unwind: true, force_no_backtrace: false }

daninskip:
Update: If I download a new instance of Arnis each time, its fine, if I try to reuse it a second time, the crashes appear.

randomlettersForS:
For me overwriting the "Arnis World 1" also works

edit; it came out that it is also one time use

louis-e:
Hi there, thanks for the report! A temporary workaround is deleting the arnis-tile-cache/ folder which is in the directory where you ran Arnis or disabling terrain. I'll include a fix for this issue in the next release!

PR #556, opened 2025-09-03
Implement faulty elevation tile cache detection crash

louis-e:
Fixes https://github.com/louis-e/arnis/issues/537


louis-e:
retrigger-benchmark

PR #563, opened 2025-09-07
Metadata

scd31:
[Note: please merge my other PRs before this one!]

Adds a JSON file with world metadata. Right now I'm just saving the world coords (MC and Geo) but I can envision more stuff being added in the future.

I threw this together as a PoC: https://arnis-map.scd31.com/ allows you to click on a part of the map (once you load the metadata.json file) and get the MC coords, plus the command to teleport there. I could also see this being used by e.g. server plugins. Eventually I want to write one that let's you do e.g. `/geotp "24 sussex drive"` and it uses the metadata to know where to teleport you.


louis-e:
retrigger-benchmark

louis-e:
LGTM
I will just improve the error handling in the metadata creation step by replacing .expect() with a more greaceful handling which otherwise would panic the program on failure, since the metadata saving isn't a crucial part of the generation pipeline. Let me know if you don't agree with that!

scd31:
> LGTM I will just improve the error handling in the metadata creation step by replacing .expect() with a more greaceful handling which otherwise would panic the program on failure, since the metadata saving isn't a crucial part of the generation pipeline. Let me know if you don't agree with that!

This probably isn't worth bike-shedding on two much, but my thinking is basically that it should be exceedingly rare for this step to fail but the other file saving steps (regions etc.) to succeed. Furthermore it may be confusing or frustrating for users if this step happens to fail and then, months down the line, they then need this file (like for one of the many arnis utilities that I expect to show up over time (: ). In that case it may be better to fail early, even if it requires them to regenerate their world

Like I said though, this isn't a huge deal either way and I don't feel strongly, just explaining my initial logic

PR #566, opened 2025-09-13
Clean up root directory

louis-e:
As of right now, the problem we face is that the root directory has quite a lot of files, leading to users having to scroll down quite a bit to view the README with valuable information on the Github page. This PR cleans up the folder structure a bit.


src/map_transformation/transform_map.rs
```diff
@@ -14,7 +14,7 @@ pub fn transform_map(
     println!("{} Transforming map...", "[4/7]".bold());
     emit_gui_progress_update(20.0, "Transforming map...");
 
-    match fs::read_to_string("example_transformations.json") {
+    match fs::read_to_string("tests/map_transformation/example_transformations.json") {
```
scd31: FYI - could use `include_str!` here so that this gets compiled into the binary

But also, would you be explain to explain the purpose of this? I don't understand what it's used for

README.md
```diff
@@ -51,7 +51,7 @@ After your pull request was merged, I will take care of regularly creating updat
 
 ## :newspaper: Academic & Press Recognition
 
-<img src="https://github.com/louis-e/arnis/blob/main/gitassets/recognition.png?raw=true" width="100%" alt="Banner">
+<img src="https://github.com/louis-e/arnis/blob/main/assets/gitassets/recognition.png?raw=true" width="100%" alt="Banner">
```
scd31: FYI, we can link to images through relative paths in GitHub: https://stackoverflow.com/questions/15764242/is-it-possible-to-make-relative-link-to-image-in-a-markdown-file-in-a-gist#36313391

I think this provides the following advantages:
- When changing assets in a branch, they show up properly in that README (instead of always linking to the main branch)
- Assets will load locally when viewing the markdown without needing to call out to GitHub (useful when not having internet access, or when GitHub is down, since it goes down pretty often)

src/map_transformation/transform_map.rs
louis-e: This was introduced by XianlinSheng in https://github.com/louis-e/arnis/pull/409. It's not used actively in the processing pipeline yet, but his implementation enables some transformations later on.

louis-e:
Thanks for the review! :)

PR #570, opened 2025-09-13
Street signs, namespaced blocks

jcobol:
Refactor Block to use string names; add road sign placement

- Replace Block ID (u8) with namespaced string names for better extensibility and readability. Add Block::from_str with global cache for dynamic blocks. Update all constants and methods accordingly.
- Add format_sign_text to wrap text for Minecraft signs (max 15 chars/line, truncate after 4 lines).
- In highways: Place signs along named roads every ~200m (scaled), with fallback for short roads. Add tests.
- In world_editor: Update set_sign to use absolute Y, add back_text, explicit properties, and force placement. Normalize palette names. Update NBT serialization with DataVersion, lights, and full root structure.
- Minor: Use fs2 for file unlock in gui; add tests for formatting and serialization.

Disclosure: used AI to build this patch

### DIFF
--- a/src/block_definitions.rs
+++ b/src/block_definitions.rs
@@ -4,6 +4,7 @@ use fastnbt::Value;
 use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+use std::sync::Mutex;
 
 use crate::colors::RGBTuple;
 
@@ -57,7 +58,7 @@ type ColorBlockMapping = (ColorTuple, BlockOptions);
 
 #[derive(Copy, Clone, PartialEq, Eq, Ord, PartialOrd, Hash, Debug)]
 pub struct Block {
-    id: u8,
+    name: &'static str,
 }
 
 // Extended block with dynamic properties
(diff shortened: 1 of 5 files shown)


louis-e:
retrigger-benchmark

jcobol:
Thanks for the feedback - I'll work on improving the patch.

jcobol:
retrigger-benchmark

jcobol:
retrigger-benchmark

louis-e:
Don't worry too much, there might be a bug in the benchmarking bot currently. Btw thanks for the contribution and the AI disclaimer, I'll review it soon!

src/floodfill.rs
```diff
@@ -160,7 +160,7 @@ fn original_flood_fill_area(
     for z in (min_z..=max_z).step_by(step_z as usize) {
         for x in (min_x..=max_x).step_by(step_x as usize) {
             // Reduced timeout checking frequency for better performance
-            if global_visited.len() % 200 == 0 {
+            if global_visited.len().is_multiple_of(200) {
```
louis-e: https://github.com/rust-lang/rust/issues/128101
use of unstable library feature 'unsigned_is_multiple_of'

src/element_processing/highways.rs
```diff
@@ -451,6 +452,92 @@ fn generate_highways_internal(
...
             }
+
+            if let Some(name) = element.tags().get("name") {
+                if debug {
+                    eprintln!("Processing highway '{name}'");
+                }
+                let mut prev_node: Option<&crate::osm_parser::ProcessedNode> = None;
+                let sign_interval = (200.0 * args.scale).max(1.0);
+                if debug {
+                    eprintln!("  Sign placement interval: {sign_interval}");
```
louis-e: Unnecessary debug statements

src/element_processing/highways.rs
```diff
@@ -451,6 +452,92 @@ fn generate_highways_internal(
                 }
                 previous_node = Some((node.x, node.z));
             }
+
+            if let Some(name) = element.tags().get("name") {
```
louis-e, line 456: image
The current sign placing logic places too many signs. I'll disable the automatic sign placing code temporarily until we found a better logic.

src/world_editor.rs
```diff
@@ -16,6 +16,62 @@ use std::io::Write;
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicU64, Ordering};
 
+const DATA_VERSION: i32 = 3700;
```
louis-e, line 19: Do you know where the constant comes from? Is this constant bound to a specific version of Minecraft?

src/block_definitions.rs
```diff
@@ -468,27 +149,43 @@ impl Block {
...
-use std::sync::Mutex;
+// Cache for dynamically created blocks by string name
+static BLOCK_NAME_CACHE: Lazy<Mutex<HashMap<String, Block>>> =
+    Lazy::new(|| Mutex::new(HashMap::new()));
 
+impl Block {
+    /// Construct a `Block` from an arbitrary namespaced string.
+    ///
+    /// The string is stored in a global cache to obtain a `'static` lifetime.
+    pub fn from_str(name: &str) -> Block {
```
louis-e, line 160: Intentionally leaks memory with Box::leak() to create 'static references. No mechanism to clean up these leaked strings during program lifetime. Could accumulate significant memory usage with dynamic block creation?

src/element_processing/highways.rs
```diff
@@ -636,3 +723,130 @@ pub fn generate_aeroway(editor: &mut WorldEditor, way: &ProcessedWay, args: &Arg
...
+        let mut tags = HashMap::new();
+        tags.insert("highway".to_string(), "primary".to_string());
+        tags.insert("name".to_string(), "First St.".to_string());
+        let way = ProcessedWay { id: 1, nodes, tags };
+        let element = ProcessedElement::Way(way);
+
+        let all = [element.clone()];
+        generate_highways(&mut editor, &element, &args, &all);
+
+        for x in [200, 400, 600, 800, 1000] {
```
louis-e, line 766: Test assumes signs are placed at specific coordinates without accounting for the actual road geometry. Test may be brittle and fail if road placement algorithm changes slightly

louis-e:
I already addressed three of my comments with commits, however there are some things we need to clarify before merging to avoid introducing regressions :)

src/world_editor.rs
jcobol: Yes, it's the Java Edition data value: https://minecraft.fandom.com/wiki/Java_Edition_data_values/Protocol_and_data_versions
3700 = [Java Edition 1.20.4](https://minecraft.fandom.com/wiki/Java_Edition_1.20.4)

src/block_definitions.rs
jcobol: I tried setting the AI coder (codex web) to get rid of this, but after 3 tries with timeouts (it gave up), I asked for the pros and cons of removing the Box::leak(). What follows is the response:
---
Switching Block to store a Cow<'static, str> would eliminate the need to leak owned strings in Block::from_str and could make it easier to accept dynamically generated block identifiers without synchronizing through the global cache. However, the rest of the crate treats Block as an inexpensive, trivially copyable value. That assumption is baked into large swaths of the codebase: arrays of blocks are copied out by indexing, helper tables are repeated with [BLOCK; N] initializers, and world-editing data structures iterate over [Block; 4096] palettes by value.

Pros
Removes the lifetime and leaking workaround in Block::from_str, letting owned names stay owned and potentially collapsing the mutex-protected cache entirely.

Gives future features (e.g., importing user-defined blocks) a natural place to store owned identifiers instead of forcing everything through 'static literals.

Cons
Block losing Copy would ripple through all array-based helpers, requiring clone()/cloned() at each random-selection site (WINDOW_VARIATIONS, get_random_floor_block, agricultural crop picks, etc.).

Core world-editing structures rely on copy semantics for performance: initializing [AIR; 4096], iterating with for (i, &block), and hashing (Block, Option<Value>) tuples all assume cheap copies; rewriting them to clone would add both code churn and per-block overhead.

Every cache lookup like *block in Block::from_str or DashMap keys would need explicit cloning, touching concurrency-critical paths and risking regressions without clear functional gain.

Recommendation
Unless there is a pressing requirement for many distinct, owned block identifiers, the maintenance and performance cost of de-Copying Block seems to outweigh the benefit. The existing cache already bounds the leak to at most one allocation per distinct name, so the pragmatic risk is modest compared to the extensive refactoring this change would force across generation and world-editing code.

PR #571, opened 2025-09-14
Add rotate-and-expand with GUI and CLI

13raur0:
This PR adds support for rotating the world before generation.  
The purpose is to allow users to specify a rotation angle to fix jagged map features.  
For example, in Manhattan, most roads and buildings appear jagged because the city is tilted on the map.  
Although a similar feature has been proposed before, I am sharing my code for reference.  
This is mostly experimental; feedback is welcome.

Note: "expand" here means adding padding to fit all rotated coordinates, not scaling the map itself.

- GUI: added rotation angle input (form + slider, synchronized, up to 0.01° precision)
- Map: added Polyline drawing button to measure angles between two points
- generate_world: apply rotation via rotate_and_expand
  - allow editing elevation_data in ground
  - expand bbox to fit all rotated coordinates (may increase chunk count)
  - padding generated at height -62
- CLI: added angle option (--rotate-angle)

I’m a fan of Arnis and look forward to its future development!

anglecalc
setangle
</br>
gen_world
</br>


XianlinSheng:
Your GUI poly line part is unique and seems very useful, but in the core of rotation there are some differences with my implementation. I will open my version and try to explain the reason of every part of my code. You can do something similar. Let's make sure we didn't miss something important and try to combine the useful parts.

XianlinSheng (commented):
I want to make sure some details. The new xzbbox seems to be the circumscribed rectangle of the rotated one, because it must be xz aligned. How is it guaranteed that the .contains(point) is inside the polygon? I introduced the contains() function because it seems that not all fetched elements are exactly inside the selected region, but including all elements that are partially inside. Did you encounter this situation? I ran the test to check the fetched data long time ago and am not sure what it is done now.

13raur0:
Thank you for checking the PR!
I’m not very confident in my English, so I apologize if I don’t explain this clearly.

In my implementation,
Just before generate_world, I apply the rotation to the coordinates in parsed_elements, xzbbox, and ground. The xzbbox is expanded to a circumscribed rectangle to ensure all elements fit within the generation area.

Elements outside the original xzbbox range are never added, and partially contained elements remain unchanged.

louis-e:
retrigger-benchmark

XianlinSheng:
you may try cargo fmt and cargo clippy to clean up the code. I'll try to run some local tests and make sure the differences in our codes pose no problems in the understanding of the data structure.

XianlinSheng:
I did some tests on this PR and mine. First of all, I appreciate that this reminded me some details about transforms on relations that I forgot. Secondly, I found out that either the API or processed elements seem to behave a little different from what I tested couple of months ago, about whether the API will truncate the way elements at the edge of bounding box. From my previous tests and knowledge on this API, it should return complete relations and ways, but now both PR showed that the elements are truncated at the bbox edge. Truncation reduces half of the workload on transformation because XZBBox has less requirements but might introduce more complexity for batching algorithms. Either case, I will say that let's make this understanding clear first.

louis-e:
Hi there, thanks for the great discussion! I'm really busy with work at the moment, but I promise I will review all the open pull requests as soon as possible. Sorry for the delay! :)

src/main.rs
```diff
@@ -114,6 +117,9 @@ fn run_cli() {
     // Transform map (parsed_elements). Operations are defined in a json file
     map_transformation::transform_map(&mut parsed_elements, &mut xzbbox, &mut ground);
 
+    //rotate_and_expand
+    let _ = rotate_and_expand::rotate_and_expand_world_data(&mut parsed_elements,&mut xzbbox,&mut ground, args.rotation_angle);
```
louis-e, line 121: Should we add a validator (or clamp) on Args::rotation_angle? GUI clamps to 0–90; CLI should match

src/gui/js/bbox.js
```diff
@@ -689,6 +731,26 @@ $(document).ready(function () {
...
+
+        // Add polyline temporarily to calculate angle
+        lineItems.addLayer(layer);
+
+        const latlngs = layer.getLatLngs();
+        if (latlngs.length === 2) {
+            const angle = calculateAngleGeo(latlngs[0], latlngs[1]);
+            const rotation = getRotationToNearestAxis(angle);
+
+            alert('Rotation for input (counterclockwise): ' + rotation.toFixed(2) + '°');
```
louis-e, line 746: Can we find a better user feedback than using alert?

src/gui/index.html
```diff
@@ -128,6 +128,15 @@ <h2 data-localize="customization_settings">Customization Settings</h2>
           </div>
         </div>
 
+        <!-- Rotation Angle Input -->
+        <div class="settings-row">
+          <label for="rotation-angle-slider" data-localize="rotation_angle">Rotation Angle (°)</label>
```
louis-e, line 133: Does the data-localize tag cause troubles during a GUI language switch if the tag has no translations in the yaml files? If yes we can just add the tag later again!

louis-e (commented):
Thanks for adding GUI/CLI rotation, this is super useful. I can also see this in the release with an automatic building alignment detection which we could add later. I have a few code review remarks and I'd like to have more of @XianlinSheng's input to decide what can be merged of both PRs to get the best out of it, before we can merge. PR #572 introduces a more general rotation operator with polygonal bboxes. I’d love to reuse that core and keep this PR focused on the GUI/CLI plumbing. But I want to take both of your and @XianlinSheng's input into account for this decision :)

13raur0:
Thanks for the feedback!
My intention with this PR was mostly to propose one possible approach to handle rotation.
I’m totally fine if parts of it get adapted, combined with #572, or modified as needed.
Please feel free to take whatever is useful from this PR.

louis-e:
Thanks a lot for your contribution, highly appreciated. Since I got multiple PRs regarding the rotation feature, I gathered the best things from all of them and merged it in https://github.com/louis-e/arnis/pull/876 :)

PR #580, opened 2025-09-26
fix world format

scd31:
The world output wasn't compliant - it would work in MC itself, but not in some utilities. I've fixed the world format and confirmed it still works in MC and Dynmap. It's a canonical 1.17 file now, so we haven't broken backwards compatibility either.


scd31:
Actually, do not merge yet. This has broken things

PR #586, opened 2025-09-28
Fix rivers

scd31:


src/osm_parser.rs
```diff
@@ -317,287 +307,3 @@ pub fn get_priority(element: &ProcessedElement) -> usize {
     // Return a default priority if none of the tags match
     PRIORITY_ORDER.len()
 }
-
-/// Clips a way to the bounding box boundaries using Sutherland-Hodgman algorithm for polygons
-/// or simple line clipping for polylines
-fn clip_way_to_bbox(
```
scd31, line 323: Bug #1 is in here somewhere - we end up chopping things off incorrectly. I've removed this code as it seems like a lot of logic that shouldn't actually be doing much (generally our element processing clamps things to within bounds anyway, right?). But I also understand someone spent a lot of time writing this in the first place. I just don't personally want to be the one to figure out what is wrong in here (plus, there may be more bugs lurking here)

src/element_processing/water_areas.rs
```diff
@@ -36,70 +36,32 @@ pub fn generate_water_areas(editor: &mut WorldEditor, element: &ProcessedRelatio
         }
     }
 
-    // Process each outer polygon individually
-    for (i, outer_nodes) in outers.iter().enumerate() {
```
scd31, line 40: I pretty much fully reverted this. Bug #2 is that processing each polygon individually completely ruins the point of the `merge_loopy_loops` function

src/osm_parser.rs
```diff
@@ -227,26 +222,20 @@ pub fn parse_osm_data(
             for &node_id in node_ids {
                 if let Some(node) = nodes_map.get(&node_id) {
                     nodes.push(node.clone());
+                } else {
+                    println!("bro what");
                 }
             }
         }
 
-        if !nodes.is_empty() {
```
scd31, line 234: These various changes is so we don't accidentally exclude a node/way that's being used in our relation

src/element_processing/water_areas.rs
```diff
@@ -36,70 +36,32 @@ pub fn generate_water_areas(editor: &mut WorldEditor, element: &ProcessedRelatio
...
-        }
+    merge_loopy_loops(&mut outers);
+    if !verify_loopy_loops(&outers) {
+        println!("Skipping relation {} due to invalid polygon", element.id);
+        return;
+    }
 
-        merge_loopy_loops(&mut inners);
-        if !verify_loopy_loops(&inners) {
-            // If inners are invalid, process outer without inners
```
scd31, line 55: This also didn't seem like a great fallback to me. If the inner loop is invalid we probably don't want to just fill it in!

scd31:
image

sick, this has also improved new york massively. It's moist!

scd31:
image

Fixed docks as well in the latest commit. https://wiki.openstreetmap.org/wiki/Key:waterway Apparently "waterway=dock" represents a water area, so I've implemented that

scd31:
image

new york city is now fully wet! badda bing badda boom

louis-e:
Incredible work as always. Tons of thanks to you!
However I'm getting those errors in Minecraft 1.21.8 when loading the world:
[20:36:16] [Worker-Main-25/ERROR]: Chunk file at [16, 17] is missing level data, skipping
[20:36:16] [Worker-Main-24/ERROR]: Chunk file at [15, 19] is missing level data, skipping
[20:36:16] [Worker-Main-23/ERROR]: Chunk file at [15, 18] is missing level data, skipping
[20:36:16] [Worker-Main-20/ERROR]: Chunk file at [17, 18] is missing level data, skipping
...

Minecraft then replaces the faulty chunks with the standard generated flat Minecraft "terrain". I fixed it locally to try out the waterway generation. Could you check back on your end if this regression is introduced by this PR?

scd31:
@louis-e Yeah that's a regression introduced by my other PR. I'll remove the commit from this so it's not blocking

scd31:
@louis-e done

louis-e:
Perfect, thank you so much!

PR #648, opened 2025-12-01
Include map preview in CI benchmark

louis-e:


louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

PR #662, opened 2025-12-07
set generate_features to 1 in level.dat nbt template

amir16yp:
manually edited level.dat template to enable feature generation, apparently, players who tried to enter the nether or end in an arnis generated world were unable to find any structures, this fixes the issue. worry not, it shouldnt generate structures on top of chunks that arnis generates


louis-e:
Thanks a lot, LGTM! :)

PR #782, opened 2026-03-14
Add void world generation option

0xExal:
Hello,

This PR adds a void world option so chunks outside the generation bounding box stay empty instead of being filled with superflat terrain.

Available in the CLI (--void-world) and the GUI.


Crinfarr:
I feel like changing the api for `world_utils::create_new_world` is overkill here, maybe something like `world_utils::create_void_world` that calls `create_new_world` internally would be more appropriate

PR #793, opened 2026-03-16
memory-fix

pingpongmury:
save chunks as they're completed rather than keeping them in memory and saving them at the end of ground generation. Results in far lower memory footprint and enables generating much larger areas.


pingpongmury:
retrigger-benchmark

louis-e:
Hey there, thanks a lot for the contribution! Looks good, I just need to fix a few things. Will merge soon

pingpongmury:
Looks like this turned into a bit of a mess - my apologies. I actually thought I was making the PR to the main branch of my fork so I could make use of your benchmarking action.

During my testing, my RAM usage dropped from more than the 32GB in my system (causing a hang and crash) to about 8-9GB during ground generation. I still saw peaks of 24+GB while processing terrain, though. I was able to generate a roughly 350sq. mile area without issue (at least that I noticed in-game).

Regardless, I appreciate you taking a look! We just got my dad into Minecraft and I used Arnis to generate his entire home-county. I showed him mountains in Colorado, and let him explore NYC. It has been blowing his mind, ha. You guys have built a supremely awesome tool and I sincerely appreciate your work.

PR #807, opened 2026-03-21
fix(railways): respect layer tag and use ascending rails on terrain slopes

### LINKED ISSUE #331: [BUG] Railways issue with elevation enabled
**Describe the bug**
If there is a railway and an elevation difference, the tracks on different levels don't appear connected together.

~~P.S. Also I think that it will look much better if the iron blocks are replaced with gravel, or coarse dirt~~ done already 

**Used bbox area**
48.431934 35.059004 48.443608 35.066214

Arnis 2.2.0, Minecraft 1.21.4, Terrain enabled

**Screenshots**
Image
Image

wabalabudabdab:
## Problem

Two related issues in `generate_railways` (`src/element_processing/railways.rs`):

### 1. OSM `layer` tag was ignored

Railways tagged with `layer=1`+ (bridges, overpasses) were always placed at `y=0`/`y=1` — rendering at ground level instead of floating above it.

### 2. No ascending rail blocks on terrain slopes

With terrain enabled, flat rail variants were placed at every point even when the ground changed height between consecutive nodes, leaving visible gaps.

Fixes #331

---

### Layer-based elevation
Read the `layer` tag and multiply by `LAYER_HEIGHT_STEP` (6 — same constant used by `highways.rs`):

```rust
const LAYER_HEIGHT_STEP: i32 = 6;
let layer_offset = layer_value * LAYER_HEIGHT_STEP;
editor.set_block(GRAVEL, bx, layer_offset, bz, None, None);
editor.set_block(rail_block, bx, layer_offset + 1, bz, None, None);
```

### Ascending rails on terrain slopes
Compare `get_ground_level()` at prev/current/next points. When current is lower than a neighbour, place `RAIL_ASCENDING_*` pointing toward the higher end:

```rust
if next_ground > current_ground {
    ascending_toward(current, next_xz)
} else if prev_ground > current_ground {
    ascending_toward(current, prev_xz)
}
```

### Vertical gap fill
When terrain rises steeply, fill the gravel substrate with `set_block_absolute` so the track always has a solid block beneath it.

---

## Changes

- `src/element_processing/railways.rs` — layer offset, slope detection, ascending rail selection, gap fill

---

## Testing

Built and ran locally against two real bounding boxes:

**1. Original bug report bbox (Dnipro, Ukraine — where #331 was reproduced)**
```
48.431934,35.059004,48.443608,35.066214
```
Terrain range: 72.2m. 13 railway ways, all at-grade. World generates cleanly; slope detection fires on the inclined sections.

**2. South-East London, UK**
```
51.4450,-0.0100,51.4550,0.0100
```
Contains **19 railway segments with `layer=1`** (bridges over the Thames and street crossings). Before this fix they rendered at ground level; after, they are placed 6 blocks above surface (matching highway overpass behavior).

Both worlds generated to completion with no panics or errors.

### DIFF
--- a/src/element_processing/railways.rs
+++ b/src/element_processing/railways.rs
@@ -3,6 +3,9 @@ use crate::bresenham::bresenham_line;
 use crate::osm_parser::ProcessedWay;
 use crate::world_editor::WorldEditor;
 
+/// Number of blocks per OSM layer level (matches highway elevation step).
+const LAYER_HEIGHT_STEP: i32 = 6;
+
 pub fn generate_railways(editor: &mut WorldEditor, element: &ProcessedWay) {
     if let Some(railway_type) = element.tags.get("railway") {
         if [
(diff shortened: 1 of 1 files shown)


louis-e:
LGTM, thanks!

PR #839, opened 2026-03-24
Update leisure.rs

BOBJOEx11g:
add mulch => podzol and earth and ground also will create dirt

little help from AI doing this


louis-e:
Thanks, merged!

PR #843, opened 2026-03-26
Add Japanese locale and language option

AI-Avalon:
Add a new Japanese localization file (src/gui/locales/ja.json) containing translated UI strings (labels, placeholders, errors and messages), and add a "ja / 日本語" option to the language selector in src/gui/index.html so the UI can be switched to Japanese.


louis-e:
Thanks a lot, LGTM! :)

PR #848, opened 2026-03-26
refactor: extract ground generation into dedicated module

louis-e:
Move the ~670-line ground layer generation loop (surface blocks, vegetation, shorelines, underground fill) from data_processing.rs into a new ground_generation.rs module. This keeps data_processing focused on OSM element orchestration while ground_generation owns the terrain pass.
Remove flushing since it won't prevent OOM crashes due to still existing memory spikes and massive generation time increase.

### DIFF
--- a/src/ground_generation.rs
+++ b/src/ground_generation.rs
@@ -36,8 +36,8 @@ use rand::Rng;
 /// Generate the ground layer for the entire bounding box.
 ///
 /// This must be called after all OSM element processing is complete and the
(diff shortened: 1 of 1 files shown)


louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

PR #849, opened 2026-03-26
fix: prevent flushed regions from being overwritten by cross-region w…

louis-e:
…rites

Two fixes for the incremental region flushing introduced in PR #793:

1. Delay flush by one chunk past region boundaries so tree canopies (up to 3 blocks spill radius) are fully written before serialization.

2. Track flushed regions and purge any stale re-created entries at save time, preventing near-empty regions from overwriting correct .mca files.

No hot-path changes — the tracking is a single HashSet insert per flush, and purge runs once at save time.

### DIFF
--- a/src/data_processing.rs
+++ b/src/data_processing.rs
@@ -934,30 +934,42 @@ pub fn generate_world_with_options(
             }
         }
 
-        // Flush any region columns that are now fully written.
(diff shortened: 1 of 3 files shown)


louis-e:
retrigger-benchmark

PR #857, opened 2026-03-28
feature: Taller worlds support, WIP.

scrdest:
**fix**: Invalid outliers are no longer throwing off the result due to being counted for min/max height aggregations 
**feature**: Adds a ref to Ground to Landuse/Natural processors to expose height data to the generator logic 
**refactor**: MIN_Y/MAX_Y are now exposed as common crate constants instead of being redefined in multiple places 
**tweak**: Deciduous trees should not generate at mountain altitudes by default 
**tweak**: Default min/max levels are set to 4k heigh worlds (WIP, will likely need a revert)

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -37,7 +37,7 @@ pub struct Args {
     pub scale: f64,
 
     /// Ground level to use in the Minecraft world
-    #[arg(long, default_value_t = -62)]
+    #[arg(long, default_value_t = -2030)]
     pub ground_level: i32,
 
     /// Enable terrain (optional)
--- a/src/data_processing.rs
+++ b/src/data_processing.rs
@@ -17,6 +17,8 @@ use std::collections::HashSet;
 use std::path::PathBuf;
 use std::sync::Arc;
 
+pub const MIN_Y: i32 = -2030;
+
 /// Generation options that can be passed separately from CLI Args
 #[derive(Clone)]
 pub struct GenerationOptions {
@@ -155,6 +157,7 @@ pub fn generate_world_with_options(
                         args,
                         &flood_fill_cache,
                         &building_footprints,
+                        &ground,
                     );
                 } else if way.tags.contains_key("natural") {
                     natural::generate_natural(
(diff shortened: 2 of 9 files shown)


scrdest:
This is currently a dirty dump of my assorted experiments to generate better mountains in Arnis - the more controversial tweaks like the default height values will be cleaned up (configified or removed).

louis-e:
Very cool, thanks! Can't wait to see the finished result :)

louis-e:
I'm quite under time pressure to release a new update, and I promised to deliver an option in the settings for no MAX_Y build height. I'll push a tweak for that on my own, feel free to refactor it if required. Just FYI :)

scrdest:
@louis-e sure thing! The current work around MAX_Y in this branch is frankly quite crude as I initially was just messing around seeing if I can get this working for my own use, so your tweak might be cleaner; I'll work around it one way or another.

I've actually had some sample screenshots ready for like 3 weeks now - I meant to add them to the PR, but got sidetracked (in no small part, by another, far more substantial PR I might have cooking, involving additional integrations... 👀). 

There ya go (true scale, Voxy for LODs, Eclipse shaders):

scrdest:
You might know this already, but for MAX_Y, this doesn't work out of the box with stock (Java) Minecraft. The blocks past whatever the build height is set to simply refuse to load in.

The fix is pretty simple - if the save has a datapack ZIP that includes an `overworld.json` setting `{"height": 4064, "min_y": -2032}` set, it will load the blocks in with no 'proper' mods required.

Right now I've been fixing this up by copying a datapack into each save manually; in theory we could just generate this programmatically and slap it into savefiles straight from Arnis for less hassle, but it's also a non-obvious feature (and idk how Bedrock handles things).

louis-e:
Thanks a lot, @scrdest!
I picked it up in https://github.com/louis-e/arnis/pull/976. I made sure to add you as a co-author!

scrdest:
> Thanks a lot, @scrdest! I picked it up in #976. I made sure to add you as a co-author!

Thank you @louis-e - appreciate the consideration! 😄

Issue #656, opened 2025-12-07
[BUG] Trees generate inside buildings


Macintosh-Fan:
Description: as the title says
BBOX area: 48.451749 -123.338356 48.488169 -123.289604
Versions: Arnis 2.4.0 (macOS), MC 1.21.0.
Screenshots:
<img width="1194" height="760" alt="Image" src="https://github.com/user-attachments/assets/6dcac2d5-a3af-4344-9b0e-575f1b47ed6f" />
To get there, run `/tp 1596.45 -19.19 2808.60`.

JasonOA888:
I checked the code. The tree generation already has building footprint checking:

**`src/element_processing/tree.rs:181-184`:**
```rust
if let Some(footprints) = building_footprints {
    if footprints.contains(x, z) {
        return;  // Skip if inside building
    }
}
```

And `Tree::create` calls in `landuse.rs` pass `building_footprints` correctly.

**Possible causes:**

1. `BuildingFootprintBitmap` may not include all building footprints (e.g., complex shapes, overhangs)
2. Trees might be generated before all buildings are processed in the pipeline
3. The bbox area in your report (Victoria, BC) might have complex building geometries

This needs deeper investigation into the processing order and footprint collection logic.

louis-e:
Should be fixed by now!

PR #861, opened 2026-03-31
Prevent tree canopies from clipping into buildings

xCipHanD:
## Summary

This PR fixes a collision bug where tree canopies could extend into building interiors, as reported in #656.

Previously, tree generation only checked whether the trunk base `(x, z)` was inside a building footprint. That prevented trees from spawning fully inside buildings, but it did not prevent leaves from expanding horizontally into nearby structures.

This change keeps the existing fast path for normal tree generation, and adds a guarded slow path only when a canopy may intersect a building footprint.

### Before
- Trees were skipped only if the trunk base was inside a building footprint
- Leaves could still extend into nearby buildings
- This was most visible when a tree spawned just outside a wall

image

### After
- Trees still use the normal fast path when no nearby building collision is possible
- If the canopy may overlap a building footprint, leaf placement is checked per block
- Leaves no longer enter building interiors through walls

image

## Root Cause

The tree generation logic already had a building footprint check for the trunk base:

`src/element_processing/tree.rs`

But leaf placement used unconditional canopy fill and round-pattern placement, so the trunk could be outside a building while the canopy still crossed into it.

### Fast / slow path canopy placement
- Added a canopy proximity pre-check using a local maximum canopy radius
- Trees that are clearly far from buildings keep the original fast placement path
- Trees near buildings switch to a slower but safe path that checks footprint overlap per leaf block

### Scope of fix
- Leaf column placement is now collision-aware when needed
- Round-pattern canopy placement is now collision-aware when needed
- Trunk placement logic is unchanged

### Build validation
- [x] `cargo check`

### Manual validation
- [x] Tree outside building still generates normally
- [x] Tree trunk inside building footprint is skipped
- [x] Tree canopy near a building no longer places leaves into interior space
- [x] Non-adjacent trees still use the original fast path

### DIFF
--- a/src/element_processing/tree.rs
+++ b/src/element_processing/tree.rs
@@ -112,6 +112,9 @@ const ACACIA_LEAVES_FILL: [(Coord, Coord); 5] = [
 
 //////////////////////////////////////////////////
 
+/// Maximum horizontal canopy radius used by the predefined tree patterns.
+const MAX_CANOPY_RADIUS: i32 = 3;
+
 /// Helper function to set blocks in various patterns.
 fn round(editor: &mut WorldEditor, material: Block, (x, y, z): Coord, block_pattern: &[Coord]) {
     for (i, j, k) in block_pattern {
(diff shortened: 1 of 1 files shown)


louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

louis-e:
LGTM, thanks!

PR #877, opened 2026-04-04
Don't generate zebra crossings for unmarked crossings

andersmmg:
Only generate zebra crossing markings on marked pedestrian crossings. Previously, crosswalks were placed on all footway crossings, including unmarked ones. Now they respect the `crossing=no`, `crossing=unmarked`, and `crossing:markings=no` tags to skip unmarked crossings.


louis-e:
Thanks, LGTM!

PR #883, opened 2026-04-04
Switched cargo run to release from debug

Macintosh-Fan:
This allows the user to run Arnis with significant performance improvements.


goteusz-maszyk:
`--release` compilation is also significantly slower and people knowing rust also know that they can use it, I think it may be misleading and annoying for development process

louis-e:
Thanks for the contribution! I agree with my previous commenter. During development, the debug variant will be preferred in most cases.

PR #886, opened 2026-04-05
Add Overture Maps building data integration

louis-e:
Complement OSM building data with Overture Maps ML-derived footprints to fill coverage gaps in areas with sparse OSM mapping (e.g., rural Africa, parts of Asia).

Key implementation details:
- STAC collections.parquet catalog for fast partition discovery (1 req instead of 512) with geographic pre-filtering
- Sparse HTTP Range requests: only download matching Parquet row groups (~4-5 MB each) instead of full partition files (~580 MB each)
- Source-based dedup: skip buildings whose primary source is OSM
- Spatial dedup: grid-accelerated centroid-in-bbox check against existing OSM buildings
- Extract height, min_height, num_floors, roof_shape, roof_material, roof_orientation, facade_color, roof_color, building class/subtype
- Smart height filtering: omit low Overture heights so buildings use the pipeline's default, keeping visual consistency with OSM buildings

### DIFF
--- a/src/element_processing/buildings.rs
+++ b/src/element_processing/buildings.rs
@@ -4476,7 +4476,7 @@ fn generate_gabled_roof(
         } else {
             pd.dist_to_edge
         };
-        let local_boost = ((pd.local_half as f64) * 0.75).round().max(1.0) as i32;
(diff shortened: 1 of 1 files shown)


louis-e:
retrigger-benchmark

PR #895, opened 2026-04-06
Implement tunnel=building_passage support

louis-e:
Render ground-level archway openings through buildings where highway ways tagged tunnel=building_passage pass through. A pre-scan phase collects passage highway coordinates into a CoordinateBitmap, which is threaded into building generation to:

- Cut wall blocks below passage height in build_wall_ring()
- Skip ground floors and place passage ceiling lintels
- Add corridor side walls where interior meets the passage
- Skip foundation pillars, corner quoins, wall depth features, window decorations, and special doors at passage coordinates
- Suppress interior furniture placement inside the passage zone

### DIFF
--- a/src/element_processing/buildings.rs
+++ b/src/element_processing/buildings.rs
@@ -1630,6 +1630,9 @@ fn build_wall_ring(
             );
 
             for (bx, _, bz) in bresenham_points {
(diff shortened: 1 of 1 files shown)


louis-e:
retrigger-benchmark

PR #910, opened 2026-04-10
Park bench

luigi052005:
I tried to make the benches look more like actual benches.
huge_2026-04-10_22 00 30

### DIFF
--- a/src/block_definitions.rs
+++ b/src/block_definitions.rs
@@ -740,7 +740,7 @@ pub fn create_stair_with_properties(
     block_with_props
 }
 // Add half=top to make it upside-down
-pub fn top_stair(mut stair: BlockWithProperties) ->BlockWithProperties {
+pub fn top_stair(mut stair: BlockWithProperties) -> BlockWithProperties {
     if let Some(Value::Compound(ref mut map)) = stair.properties {
         map.insert("half".to_string(), Value::String("top".to_string()));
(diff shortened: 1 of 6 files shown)


luigi052005:
I think the result would be genuinely better if the bench always faced the nearest path instead of being given a random rotation. Is that something I should try to implement?

louis-e:
> I think the result would be genuinely better if the bench always faced the nearest path instead of being given a random rotation. Is that something I should try to implement?

I like the idea and I struggled with that in the past. It would make it look better visually, but I'm concerned about the impact on the generation time. Performing the calculations to correctly align the benches might be a few milliseconds on a small area, but it will add up on large areas. In the features I added in the past few months, I always tried to weigh up what matters the most for the individual cases: visual improvement or generation time impact? I'm aware that this shouldn't be too strict, so I will try to judge it on an individual basis after taking a closer look at the calculation efficiency.

luigi052005:
We can also just keep the orientation random, like before.

luigi052005:
retrigger-benchmark

louis-e:
retrigger-benchmark

louis-e:
Just had a deeper look, looks good to me. I also like the idea of overhanging traffic signals. The ↗ 119% more memory usage isn't caused by this, will investigate where this comes from. Thanks!

PR #919, opened 2026-04-12
Revert Block.id from u16 to u8 to fix ~360MB memory regression

louis-e:
PR #906 widened Block.id from u8 to u16 to accommodate 5 new wall banner block IDs (256-260). This doubled every block's storage size, causing ~360MB peak memory increase on large worlds since BlockStorage stores Vec<Block> arrays of 4096 entries per chunk section.

Fix by moving the 5 wall banner IDs into the existing gap at 145-149 (IDs were unused between oak_stairs at 144 and chest at 155), allowing Block.id to revert to u8.

Also removes 31 block constants that were defined but never referenced anywhere in the codebase.


louis-e:
retrigger-benchmark

louis-e:
retrigger-benchmark

PR #920, opened 2026-04-12
Hanging traffic signals

luigi052005:
Used the road bitmap mask to find the nearest non road block next to a traffic signal position, so the signal mast is placed on the roadside rather than in the middle of the road. When a valid anchor is found, iron bars are strung across to the crossing point to create a hanging signal.


louis-e:
LGTM, thanks!

PR #931, opened 2026-04-14
feature: add amenity:drinking_water

iluminatiftw:
This PR is to add the amenity **drinking_water** to world generation.

image

To do this i have added the block `water_cauldron` and `lever`, i have noticed that there aren't many block ids available, so if i need to use the blocks that are already added, i thought of generating the `drinking_water` as a simple cauldron, let me know what you think.

This is my first pull request, so sorry in advance.


luigi052005:
I like it

louis-e:
Very cool, thanks! LGTM

PR #939, opened 2026-04-16
Add multi-source elevation pipeline with terrain improvements

louis-e:
- Add regional elevation providers (USGS 3DEP, IGN France, IGN Spain, GSI Japan) with auto-selection by bbox overlap and resolution ranking
- Add NaN-ratio fallback: if regional provider returns >50% NaN, fall back to AWS Terrain Tiles
- Replace percentile-based outlier filter with IQR-based detection to preserve real mountain terrain
- Add slope-based block palette (deepslate/tuff/andesite/gravel on steep terrain, Tellus-inspired)
- Add adaptive terrain depth fill using 8-neighbor ground level checks to seal cliff face gaps
- Add universal depth fill that runs for ALL columns including OSM-covered ones (fixes quarry gaps)
- Fix tree leaf placement to use absolute Y from trunk base instead of per-block ground resolution
- Add rock variation for natural=bare_rock/cliff/quarry (stone/andesite/cobblestone/gravel mix)
- Add DEEPSLATE (ID 151) and TUFF (ID 152) block definitions
- Add fill_blocks_absolute to WorldEditor for absolute Y coordinate volume fills
- Fix progress bar monotonicity (values now increase consistently through pipeline stages)
- Update GUI credits to match active elevation datasets only
- Remove unused disabled provider code and CRS conversion helpers
- Pass cargo clippy --all-targets --all-features -- -D warnings with zero diagnostics


louis-e:
retrigger-benchmark

PR #958, opened 2026-04-21
Add Chinese README translation with language selector

herobrine19:


louis-e:
Hey there, I'd like to keep the root directory as clean as possible. Thus I will close this PR. However thanks for your effort!

Issue #959, opened 2026-04-21
[BUG] stuck at 3 percent and, "Error! Server overloaded. Try again."


jancarlosayala38-cloud:
why does it get stuck at 3 percent every time a make a new generation.
32.885056,-97.276736,32.886245,-97.2751
i am on v2.6.0 windows

<img width="978" height="628" alt="Image" src="https://github.com/user-attachments/assets/76cb205e-2840-4bd9-b034-d639315fc4d3" />

<img width="983" height="628" alt="Image" src="https://github.com/user-attachments/assets/22b918a4-fd90-49c0-97ca-ee602dc27a2b" />

jancarlosayala38-cloud:
please help.

jancarlosayala38-cloud:
how  do i fix this?

thesomewhatyou:
> Server overloaded

i think you know what this means. try again later. this'll usually resolve itself

SuperSeoul:
Use a VPN to fix the a bug.

QuickMash:
this seems very common

louis-e:
Hey there, the OpenStreetMap servers where we get our data from are currently overloaded. This will be fixed in the next release

Issue #955, opened 2026-04-21
[BUG] Error 406 on all primary servers


neonxsl:
**Describe the bug**
using the latest release i am consistently getting error 406 on all primary servers.. this then fallbacks to the backup servers but those are not that great...

**Used bbox area**
I purposefully tried a TINY bbox area but alas... still the same error -33.857587,151.206894,-33.846467,151.21685

**Arnis and Minecraft version**
im on macos on latest arnis version in releases

<img width="671" height="210" alt="Image" src="https://github.com/user-attachments/assets/aed8ba78-6ce7-4c1f-9fd5-a41d19889881" />

**Additional context**

Epic34-cyberdudder:
I'm on windows and its kinda similar. i get stuck at 3% then get a server overloaded error or connection timed out

neonxsl:
i think i have found the root cause.. 
the useragent that arnis uses causes this.. (thanks to codex lmao)

arnis/2.6.0 => 406
Arnis/2.6.0 (+https://github.com/louis-e/arnis) => 200
Mozilla/5.0 => 406
curl/8.7.1 => 200
arnis => 200
Arnis => 200

after patching this.. it works as normal!

neonxsl:
<img width="490" height="113" alt="Image" src="https://github.com/user-attachments/assets/4d14d3d7-2091-41f9-9df5-60ee32bc275b" />
update this in elevation_data.rs, overture.rs, retrieve_rata.rs and version_check.rs 

thanks codex :D

RFrappe:
can confirm that the linux version of Arnis does the same thing

<img width="809" height="257" alt="Image" src="https://github.com/user-attachments/assets/23311281-6485-4972-8614-1d6239c8d18d" />

jancarlosayala38-cloud:
what exactly do i need to do to fix this?

jancarlosayala38-cloud:
HOW TO FIXXX😢

BADtochka:
[...]

PR #961, opened 2026-04-21
Fix Overpass request user agent

neonxsl:
The current user agent is getting 406 error from all primary servers. Changing the user agent fixes that.


louis-e:
Perfect, thank you so much. Thinking about releasing a hotfix for this

BADtochka:
This related to #959 #955

PR #972, opened 2026-04-23
Make note about building generation

QuickMash:
Added an important note about reliance on OpenStreetMap for building data.


louis-e:
We now pull in Overture Maps data for additional building data!

Issue #853, opened 2026-03-27
Park generation should respect the given surface type


Westbrooke117:
Some parks near me are correctly tagged as having a surface of grass, with either trees individually mapped, or with natural wood areas within them. arnis seems to consider all area within the park polygon as natural wood regardless, which causes some parks which have sparse trees IRL to appear as dense forests within arnis.

**Describe the solution you'd like**
Have the program respect the surface tag given to the park (grass, concrete, etc.), and only use natural wood coverage as a fallback.

eengnr:
An example is the Olympiapark in Munich:
48.163023,11.539121,48.176446,11.560321
This grass surface has many trees in the generated map: https://www.openstreetmap.org/way/1082418004

eengnr:
There are probably two places in the code where this could be optimized.
The first is in the `leisure.rs` where trees are generated with a random chance between 105..200:
https://github.com/louis-e/arnis/blob/78235559f7cc7e02f82409c6a3f201947960a31c/src/element_processing/leisure.rs#L125
This should be removed.

The second is a little bit harder, I guess.
I found out that "forest" wins over "meadow" or "grass" if the forest surrounds a grass area.
You can see this e.g. here: 48.101814,11.901326,48.110984,11.917806
A lot of trees are generated and the meadow area https://www.openstreetmap.org/way/92836785 is "overwritten".
`forest` here should only be considered if no `grass` or `meadow` is additionally found:
https://github.com/louis-e/arnis/blob/78235559f7cc7e02f82409c6a3f201947960a31c/src/element_processing/landuse.rs#L181

PR #981, opened 2026-04-24
Fix tree generation on grass areas in forest areas

eengnr:
This is an attempt to improve the tree generation in parks, see #853 
It avoids generating trees on grass areas by sorting landuse tags with a priority.


eengnr:
It's possible that this change has the opposite effect on forest areas that are inside of grass areas... I'll check this.

eengnr:
Not necessary anymore, better solution in #985

PR #985, opened 2026-04-26
Smallest area check for landuse, leisure and natural

### LINKED ISSUE #440: [BUG] Nested landuse can lead to wrong blocks
**Describe the bug**
There are areas where e.g. a `landuse=grass` is within a `landuse=residential`.
This can lead to `STONE_BRICKS` instead of `GRASS_BLOCKS`, because the outer `residential` seems to overwrite the inner `grass`.
I guess it is necessary to determine the innermost way (area) of a point/coordinate/block to avoid this.

**Used bbox area**
48.082179 11.672072 48.086938 11.682200

Arnis: self build on Linux from https://github.com/louis-e/arnis/commit/1a0fc291b00534dacea7a3d3ca967a4088952fea
Minecraft: Server with 1.21.5 hosting the generated Java map, playing with GeyserMC with Bedrock 1.21.72

**Screenshots**

Image
The triangle within the fence should be grass.

Grass: https://www.openstreetmap.org/way/1081794163
Residential: https://www.openstreetmap.org/way/654718825

### LINKED ISSUE #853: Park generation should respect the given surface type

eengnr:
With this PR, landuse and leisure ways and relations are checked for their area size and only the smallest one is considered for a position. This avoids overwriting of tags.
Additionally the random tree generation is slightly modified to generate less trees and skip certain areas.

Fixes #440
Fixes #853

### DIFF
--- a/src/data_processing.rs
+++ b/src/data_processing.rs
@@ -6,7 +6,9 @@ use crate::floodfill_cache::FloodFillCache;
 use crate::ground::Ground;
 use crate::ground_generation;
 use crate::map_renderer;
-use crate::osm_parser::{ProcessedElement, ProcessedMemberRole, ProcessedNode, ProcessedWay, ProcessedRelation};
+use crate::osm_parser::{
+    ProcessedElement, ProcessedMemberRole, ProcessedNode, ProcessedRelation, ProcessedWay,
+};
 use crate::progress::{emit_gui_progress_update, emit_map_preview_ready, emit_show_in_folder};
 #[cfg(feature = "gui")]
 use crate::telemetry::{send_log, LogLevel};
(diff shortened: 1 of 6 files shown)


louis-e:
retrigger-benchmark

src/data_processing.rs
```diff
@@ -26,6 +28,69 @@ pub struct GenerationOptions {
...
+        }
+    }
+
+    // Merge outer ways to polygon
+    merge_way_segments(&mut outers);
+
+    // Fill every merged ring as one way
+    for ring in outers {
+        let way = ProcessedWay {
+            id: rel.id,
```
daveth3t3chg33k, line 86: Reinstantiate the variable before calling it in the var_func

PR #1000, opened 2026-04-30
refactor: split math from the rest of the codebase

goteusz-maszyk:
This PR splits pure math calculations into a separate crate to reduce compile time and improve legibility


louis-e:
Hey there, this actually seems to slow down  the generation time a bit. Will skip this for now!

PR #1015, opened 2026-05-07
Filter ignored OSM tags at parse time to reduce memory

louis-e:
Drop tags Arnis never reads (created_by, wiki*, addr:*, source*, name:*, tiger:*, NHD:*, regional import refs, etc.) when constructing ProcessedNode/Way/Relation. Audited against all tags.get/contains_key sites in src/ to confirm no read keys conflict.


louis-e:
retrigger-benchmark

PR #1025, opened 2026-05-14
Bug fixes & Material improvements

suprohub:
Added more materials.
Added more surfaces.
Added using landcover and surface tags in landuse.
Fixed pyramidal roof (now last lower layer of stairs on roof looks normal)
Fixed leisures: now they flat, even if they placed in mountains:
test
test2


louis-e:
Thanks, adapted it in https://github.com/louis-e/arnis/pull/1164 to get it into the next release!

suprohub:
> Thanks, adapted it in #1164 to get it into the next release!

oka

PR #1055, opened 2026-05-21
Bake chunk lighting so generated worlds load lit

louis-e:
Chunks were written as Status=full but without lighting, so off-disk LOD/pre-gen tools (Voxy, Chunky) rendered them dark until visited and relit in-game. Compute SkyLight (vertical cast + flood-fill) and BlockLight (flood-fill from emitters) per section and set isLightOn=1 so the light engine loads them as LIGHT_AND_DATA immediately.


louis-e:
retrigger-benchmark

PR #1067, opened 2026-05-27
perf(world_editor): single-traversal write on the default no-override path

rzmo:
`set_block(..., None, None)` is the dominant write pattern across all element processors (~240 callsites). It currently goes through `get_block()` followed by `set_block_with_properties()`, descending the region/chunk/section `FnvHashMap` hierarchy twice per write.

This PR adds `WorldToModify::set_with_props_if_absent`, a single-traversal variant that does one `entry().or_default()` descent and checks `section.storage.get(idx) == AIR` directly. The `None/None` branch of `set_block_with_properties_absolute` routes to it via an early-return. The whitelist/blacklist slow path is untouched. Same pattern as the existing `set_block_if_absent` used by `ground_generation`.

`SectionToModify::get_block` returns `None` iff `storage.get(idx) == AIR`, so the condition checked is identical to the old path. Behaviour is preserved, only the number of map lookups is halved.

Benchmarked on Munich centre (`48.125768,11.552296,48.148565,11.593838`) with `--terrain --benchmark --file` over cached OSM data, release build, 5 runs each:

| | Mean | Median |
|---|---|---|
| Before | 9925 ms | 9916 ms |
| After  | 9480 ms | 9441 ms |
| Delta  | **-445 ms (-4.5%)** | **-475 ms (-4.8%)** |

All five after-runs were faster than all five before-runs.

### DIFF
--- a/src/world_editor/common.rs
+++ b/src/world_editor/common.rs
@@ -489,6 +489,45 @@ impl WorldToModify {
         }
     }
 
+    /// Like [`set_block_if_absent`] but also persists optional NBT properties.
+    #[inline]
+    pub fn set_with_props_if_absent(
+        &mut self,
+        x: i32,
+        y: i32,
+        z: i32,
+        block_with_props: BlockWithProperties,
+    ) {
+        let chunk_x: i32 = x >> 4;
+        let chunk_z: i32 = z >> 4;
+        let region_x: i32 = chunk_x >> 5;
+        let region_z: i32 = chunk_z >> 5;
+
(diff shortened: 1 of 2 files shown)


rzmo:
The benchmark ran successfully but the verdict was "generation time is unchanged" since CI rounds to whole seconds, so the local 4.5% improvement falls below CI's resolution. Can retrigger, I don't mind

louis-e:
Perfect, thank you so much!

PR #1099, opened 2026-06-13
Update landuse.rs

BOBJOEx11g:
expanded landuse.rs to include more OSM elements, all elements have more than 3000 uses in OSM.


louis-e:
Thanks for the contribution! I stripped it down to farmyard since the other elements partially cover a much bigger area than usually.

PR #1108, opened 2026-06-16
fix(water): flatten large seas/lakes instead of terracing them (#1066)

### LINKED ISSUE #1066: [BUG] Water generation error around Monaco
**Describe the bug**
Along the coast of Monaco, parts of the Mediterranean generate at abnormal Y levels, resulting in large canyons and ravines stretching across the sea.

**Used bbox area**
43.712687,7.384443,43.765521,7.461863

Arnis v2.8.0 for Windows, Minecraft 26.1.2

**Screenshots**
Image
Image
Image

**Additional context**
This does not occur if legacy terrain is enabled or a smaller bbox is used. Interior Generation and Fill Ground were enabled.
Log file: [arnis.log](https://github.com/user-attachments/files/28306068/arnis.log)

louis-e:
level_water_surfaces classified the open sea as "flowing" whenever its interquartile elevation range exceeded 5 m. Over a large bbox the sea accumulates enough noisy AWS bathymetry to cross that threshold (Monaco: IQR 480 m), and the flowing branch's per-cell square-window median terraces it into canyons and straight lines across the water (only at large bbox, hence "not on a smaller bbox"). Force any body covering >=5% of the grid to the still branch (one histogram-mode surface); rivers cover a tiny fraction and keep their gradient. Verified on the Monaco bbox: flowing components 1 -> 0, the sea levels flat.

### DIFF
--- a/src/elevation/postprocess.rs
+++ b/src/elevation/postprocess.rs
@@ -350,7 +350,17 @@ fn level_water_surfaces(heights: &mut [Vec<f64>], lc_grid: &[Vec<u8>]) -> Vec<Ve
                 v[mid]
             };
 
-            if iqr > FLOWING_IQR_THRESHOLD_M {
+            // A body covering a large fraction of the grid is a sea/ocean/major
+            // lake: its high IQR comes from noisy bathymetry, not a real gradient,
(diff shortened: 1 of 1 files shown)


louis-e:
Needs more time to fix

PR #1120, opened 2026-06-19
Fix water areas rendering as a straight-edged wedge

Teddy563:
Some water areas were rendering as a big triangle or rectangle of water with a
straight diagonal edge cutting across the map (see screenshots).

The cause is in `clip_water_ring_to_bbox`. It expects a closed ring, but sometimes
a ring comes in open (for example when a multipolygon's member way did not load).
The code then closes it by drawing a straight line from the last point back to the
first, and that straight line gets flooded as water.

The fix just checks if the ring is actually closed before filling it. If it is not,
it gets skipped instead of being forced shut. Closed rings render exactly as before,
so you get partial water instead of a wedge. Added a few unit tests for the closed
and open cases.

## Screenshots
Left half is before, right half is after.

fixed_broken
fixed_broken2

### DIFF
--- a/src/clipping.rs
+++ b/src/clipping.rs
@@ -68,12 +68,36 @@ pub fn clip_way_to_bbox(nodes: &[ProcessedNode], xzbbox: &XZBBox) -> Vec<Process
     assign_node_ids_preserving_endpoints(nodes, polygon, way_id)
 }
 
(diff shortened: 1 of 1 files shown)


louis-e:
Thanks, I'm pretty sure I already backported your fix! :)

PR #1158, opened 2026-07-09
Mapterhorn elevation

louis-e:


deltaly23:
@louis-e why Mapterhorn is better than aws elevation?

louis-e:
It provides several datasets as one source, with a much better data resolution

PR #1167, opened 2026-07-11
Add is_empty method and tests for OsmData

devops2626:
## Summary
- Added `is_empty()` method to `OsmData` to check for empty datasets.
- Added unit tests for the new method.
- Replaced dangerous `.expect()` panics in `main.rs` with `Result` propagation.

## Why
- Prevents crashes when file operations fail.
- Makes the codebase more robust and aligns with professional Rust practices.

## Testing
- Code review only; CI will run tests automatically.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -191,7 +191,7 @@ dependencies = [
 
 [[package]]
 name = "arnis"
-version = "2.9.0"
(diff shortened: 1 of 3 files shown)


devops2626:
IMG_2358

PR #1185, opened 2026-07-12
Add cargo module with initial implementation and dependencies

devops2626:


louis-e:
This adds node_modules with the npm package "cargo", which is a JS localStorage library and has nothing to do with Rust's Cargo. Closing.

PR #1232, opened 2026-07-31
perf: reduce repeated processing work

lyshqq:
## What changed

- retain cached way flood fills until their final relation consumer finishes
- replace linear removed-index lookups during way-ring merging with constant-time markers
- cache relation member bounding boxes while assigning elements to tiles
- scan already-resolved chunk sections when locating the highest block in a column
- avoid temporary `String` allocations for several tag comparisons

## Why

These paths repeat work inside relation, tile, and world-generation hot loops. The changes reduce redundant polygon fills, node scans, hash lookups, and small heap allocations while preserving existing processing order and output semantics.

## Impact

Large or relation-heavy map generations should spend less CPU time in preprocessing and world editing. Sequential generation also retains flood-fill data only until the last relation that needs it, keeping cache lifetime bounded.

## Validation

- `cargo fmt -- --check`
- `git diff --check`
- attempted `cargo test --all-targets --all-features`; dependency compilation was blocked locally because the Visual C++ linker (`link.exe`) is not installed

A focused unit test covers the new section-wise highest-block lookup.


louis-e:
retrigger-benchmark

PR #1253, opened 2026-08-10
Fix nix build by adding missing dependency and updating lock file

KevinHermelin:
Thanks for the project! It is very cool and, as a nix user, I also appreciate the included nix flake. 

The flake fails to build right now because of two reasons:

- dda-voxelize is fetched from a Github repo but its hash is not specified in the flake.
- cfg_select is used by a dependency but is not available on rustc 1.88.0. 

To fix this, this PR does two things:

- Adds the hash of "dda-voxelize" to `flake.nix`
- Updates the `flake.lock` by running a `nix flake update`. This bumps rustc to 1.97.1.

If you are using nix, you can verify that this builds cleanly by running `nix run github:KevinHermelin/arnis -- --help`
(and compare it to upstream `nix run github:louis-e/arnis -- --help`)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,12 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1755615617,
-        "narHash": "sha256-HMwfAJBdrr8wXAkbGhtcby1zGFvs+StOp19xNsbqdOg=",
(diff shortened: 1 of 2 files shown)


louis-e:
Thank you for your contribution! LGTM :)

PR #1260, opened 2026-08-12
fix: unfreeze sun in Bedrock worlds with extend build height

### LINKED ISSUE #1254: [BUG] Sun freezes with option "Extand build height"
**Describe the bug**
Visually, time freezes in game when you create a world with the option "Extand build height". Time does pass though, as indicated on the screenshots, but the sky is frozen.

**Used bbox area**
Here's my input parameter : 43.09853,3.07492,43.115957,3.097033   , but time does not pass on any world created with this setting enabled.*

I'm using the last version of Arnis and playing on 1.21.11 on windows with fabric and mods, though i've tried without them and it still does the same.

**Screenshots**
Image

Image

Thank you for your time.

weed33834:
### Problem
When generating Bedrock worlds (especially with the "Extend build height" option), the sun visually freezes in-game. Time still passes according to the HUD, but the sky remains static.

### Root Cause
In `bedrock.rs`, `daylight_cycle` was hardcoded to `0` in the level.dat structure. In Bedrock, `daylightCycle: 0` means "always day" (frozen time), which conflicts with `do_daylight_cycle: true` also being set.

### Solution
Changed `daylight_cycle` from `0` to `1`, which enables normal day/night progression. This is consistent with `do_daylight_cycle: true` already set in the game rules.

### Testing
- Verified the level.dat field mapping matches Bedrock documentation
- `do_daylight_cycle: true` + `daylight_cycle: 1` enables normal time progression
- No impact on Java edition (this field is Bedrock-only)

Fixes #1254

### DIFF
--- a/src/world_editor/bedrock.rs
+++ b/src/world_editor/bedrock.rs
@@ -421,8 +421,8 @@ impl BedrockWriter {
             permissions_level: 0,
             player_permissions_level: 1,
 
-            // Daylight cycle
-            daylight_cycle: 0,
+            // Daylight cycle (1 = normal, 0 = frozen)
+            daylight_cycle: 1,
         };
 
         let nbt_bytes = nbtx::to_le_bytes(&level_dat)


louis-e:
Thanks for the contribution, fix however provided in https://github.com/louis-e/arnis/pull/1266

PR #1261, opened 2026-08-12
fix: reduce default height for single-family homes

### LINKED ISSUE #1239: [BUG] Buildings are way too high everywhere
**Describe the bug**
All buildings use like 5 blocks per floor, if the terrain is 1 blok  = 1 meter a house floor should not be more then 3 blocks high. ity gives a really unrealistic look to all the buildings and they all feel stretched upwards

Arnis 3.0.0
Minecraft 1.26.2 
Linux.

weed33834:
### Problem
All buildings without an explicit `building:levels` tag defaulted to 10 blocks (2 floors × 4 + 2 bonus), making houses look stretched upward. Users reported "buildings are way too high everywhere" with "5 blocks per floor".

### Root Cause
`calculate_building_height()` defaults to `2 * 4 + 2 = 10` blocks when no `building:levels` or `height` tag is present. `adjust_height_for_building_type()` only adjusted garages/sheds (2 blocks), apartments (15), and hospitals (23), but not regular houses.

### Solution
Added `house`, `detached`, `residential`, `bungalow`, and `cabin` to `adjust_height_for_building_type()`. These now default to **6 blocks** (~1.5 floors) instead of 10, giving more realistic proportions for residential areas.

Buildings with explicit `building:levels` or `height` tags are unaffected (the adjustment only applies when `building_height == default_height`).

### Testing
- Existing unit tests pass (they test `calculate_building_height`, not `adjust_height_for_building_type`)
- `house` without levels: 10 → 6 blocks ✅
- `house` with `building:levels=2`: still 10 blocks ✅ (not default, so not adjusted)
- `apartments` without levels: still 15 blocks ✅
- `garage` without levels: still 2 blocks ✅

Fixes #1239

### DIFF
--- a/src/element_processing/buildings.rs
+++ b/src/element_processing/buildings.rs
@@ -1729,6 +1729,13 @@ fn adjust_height_for_building_type(
     let default_height = ((10.0 * scale_factor) as i32).max(3);
     match building_type {
         "garage" | "garages" | "carport" | "shed" => ((2.0 * scale_factor) as i32).max(3),
+        // Single-family homes without explicit levels default to 6 blocks
+        // (~1.5 floors) instead of 10 to avoid stretched-looking houses (#1239)
+        "house" | "detached" | "residential" | "bungalow" | "cabin"
+            if building_height == default_height =>
+        {
+            ((6.0 * scale_factor) as i32).max(3)
+        }
         "apartments" if building_height == default_height => ((15.0 * scale_factor) as i32).max(3),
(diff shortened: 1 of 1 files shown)


Minty-Fresh-Github:
This is amazing! I've wanted this feature for a while!

louis-e:
Thanks for digging into this, and sorry for the collision. #1239 was fixed in the building overhaul (#1269) a couple of days ago! :)

PR #1264, opened 2026-08-13
Farmland texturing: parcels, monoculture crop plots, and a five-way preset

Teddy563:
### The problem

`landuse=farmland` renders as one uniform sheet of crops. Real farmland is a patchwork of separate plots, each worked on its own and each growing a single thing. From above, a rural area currently reads as one flat colour rather than as a field system, which stands out against how accurate the rest of the world is.

### What this does

Splits farmland into rectangular **parcels** with dirt-track boundaries and a fine internal sub-noise, and gives every farm parcel exactly **one crop**: wheat, potato, carrot, beetroot, sunflower, pumpkin or fallow. Plots carry interior character, including worn coarse-dirt spots, a mid-plot working path on large parcels, sunflower rows planted on dirt, and a pumpkin patch on a grass and coarse mosaic.

Parcel grids take their rotation from the dominant nearby road, since real fields are laid out off their access road, and fall back to one of six hashed angles where no road is close. Sunflower plots favour the low ground. That is sampled at the parcel centre rather than per block, so a plot lying across a lowland border never grows two crops.

Fallow plots deliberately use a bare palette of coarse dirt, rooted dirt and packed mud rather than farmland, because farmland with no crop above it reverts to dirt once the world is loaded in game. The same reasoning picks packed mud for the strips between sunflower rows: it stays bare instead of regrowing grass.

### A preset, not a pile of flags

One new argument, `--fields`, with a matching segmented control in the settings panel, following the same pattern as the existing Max Tree Size control:

| Preset | Look |
| --- | --- |
| `classic` | Current behaviour. The field pass does not run at all. |
| `smallholding` | Many small plots, full crop variety, dense tracks |
| `patchwork` | Balanced mixed farmland (default) |
| `prairie` | Large industrial fields, wheat-led, few tracks |
| `pasture` | Grazing land: grass and wildflowers with a few crop plots |

Each preset sets three things at once: the share of each parcel style, the parcel size band, and the crop weights. I went with a single preset rather than exposing those as separate list-valued flags, since five named looks are easier to reason about than three orthogonal knobs, and it keeps the GUI to one row.

Parcel sizes are defined in real-world metres and rescale with `--scale`, so a 1:10 world gets parcels roughly twice as many blocks across as a 1:20 world, and both cover the same amount of actual ground. There is a test pinning that.

### Seeing it

I do not have screenshots that would survive compression well enough to be useful, so here is a one-command repro on a bbox with real OSM farmland polygons:

```
cargo run --release --no-default-features -- \
  --bbox="47.400,8.500,47.420,8.530" --fields=patchwork \
  --output-dir="YOUR_PATH/saves/fields-patchwork"
```

Swapping `--fields=classic` on the same bbox gives the current output for comparison. `prairie` and `smallholding` are the clearest contrast pair if you only want to look at two.

### Determinism and seams

Everything is a pure function of `(x, z)`. No per-run state, no RNG threading, no dependence on which tile resolves a given block. The layout is therefore identical across tile boundaries and across separate runs of overlapping bounding boxes.

The two supporting lookup grids follow the same rule and each ships a test pinning it:

- `road_bearings` aggregates highway segment directions into a world-anchored 64-block lattice. The cell size is deliberately small: a domain queries the bearing at its own centre, which can sit up to 96 blocks outside the area a tile keeps, so reaching further than one cell would let two tiles with different clipped road sets resolve different rotations and visibly rotate the parcel grid at a seam.
- `lowland` samples terrain height on a world-anchored 128-block lattice and marks the low band.

Both are `div_euclid` on absolute coordinates rather than being indexed from a bbox origin, which is the part that matters.

### Verification

Checked against a pristine build of `main` on the Swiss farmland bbox above, 25600 chunks.

`main` is not fully deterministic run to run: rendering the same bbox twice with the same unmodified binary produced different block content in 4839 of those chunks. So rather than claim byte-for-byte parity, I measured only on the 20761 chunks `main` reproduces identically across two of its own runs. On that stable set, `classic` matched `main` exactly, while `patchwork` changed 1759 chunks. Comparison was done on parsed section palettes and block-state arrays, since chunk NBT tag order also varies per run.

Also: `cargo test` passes (431, of which 13 are new), `cargo clippy --all-targets -- -D warnings` is clean, and `cargo fmt --check` is clean.

On performance, I could not separate the presets from `main` above measurement noise on my machine, where repeated identical runs varied between 17 and 40 seconds. So rather than quote a figure, here is the actual work added: one parcel resolution per farmland block, which is a few integer hashes plus three value-noise samples with no allocation, and one read-lock acquisition on each of the two lookup grids. It only runs on blocks inside a `landuse=farmland` polygon, and not at all under `classic`. If you would like a proper measurement on quieter hardware, tell me which bbox you benchmark with and I will run it.

### Scope

- New `src/element_processing/field_texture.rs`, holding the parcel lattice, orientation domains, crop assignment and surface resolution
- New `src/road_bearings.rs` and `src/lowland.rs`, two small world-anchored lookup grids built once per run
- `landuse.rs`, parcel-aware surface plus per-crop decoration with growth stages
- 14 new block definitions, being 13 distinct blocks plus the sunflower's upper half
- 13 unit tests, and all 18 locale files updated in lockstep with the new UI strings

Roughly +1546 and -1 lines. Nothing existing is removed, and no existing block id is reused.

### Two things that are your call

**The default.** I have it shipping as `patchwork`, on the grounds that a feature defaulted to off is a feature nobody sees. But it does change how every existing user's farmland looks. Changing the default to `Classic` is a one-word diff and I am glad to push that instead if you would rather the merge be visually inert.

**Follow-up scope.** While testing I found that most rural farmland does not come from OSM at all. On a bbox in southern Romania with zero `landuse=farmland` ways, the ground still rendered as farmland, all of it from the ESA cropland branch in `ground_generation.rs`. Applying the same parcel system there is about 230 more lines and is what would make this visible in areas mappers have not drawn field polygons for. I kept it out so this PR stays reviewable, and can send it as a follow-up if this direction looks right to you.

Happy to split, rename, or rework any of it. Thanks for the project.


louis-e:
retrigger-benchmark

louis-e:
Rebased onto main and pushed to this branch. The only conflict was the 359-366 id renumbering.

Holding the merge for now because it increases the peak RAM memory from 702 MB to 4.3 GB, and the generation time 2.3x. I measured on a farmland dense bbox (52.520,5.600,52.550,5.660, 61 farmland polygons, ARNIS_STREAM_TO_DISK=0 to pin the code path). Since I'll try to publish a new release very soon, let's try to fix this without time pressure :)

Teddy563:
retrigger-benchmark

Teddy563:
Fixed. Reproduced your case first (same bbox, `ARNIS_STREAM_TO_DISK=0`) at 4318 MB,
which matches the 4.3 GB you measured.

### This branch, before and after

Best of three interleaved runs.

|            | before  | after   | change   |
| ---------- | ------- | ------- | -------- |
| peak       | 4318 MB | 1188 MB | **-72%** |
| time       | 30.1 s  | 21.9 s  | **-27%** |
| world hash | `b0775b84193be8d3` | `b0775b84193be8d3` | identical |

Byte identical output. Nothing about how farmland looks has changed.

### Feature cost against classic

| `--fields`  | time   | peak    | vs classic, before | vs classic, after |
| ----------- | ------ | ------- | ------------------ | ----------------- |
| `classic`   | 18.4 s | 757 MB  | baseline           | baseline          |
| `patchwork` | 21.3 s | 1151 MB | +64% time, +470% memory | **+16% time, +52% memory** |

PR #1265, opened 2026-08-14
Fix crops going dark under baked lighting, and oversized polygons rendering as a bare outline

louis-e:
Two unrelated bugs that both silently produce wrong output rather than failing. Neither is new; both pre-date v3.0.0.

## Crops go dark when baking chunk lighting

`is_light_transparent` in `src/world_editor/java.rs` lists ~48 plants and thin decor, but `wheat`, `carrots` and `potatoes` are missing, so `light_opacity` reports them as fully opaque. The skylight column scan breaks at the first opaque block:

```rust
if opacity[g] >= 15 {
    break;
}
sky[g] = level;
```

so the crop's own cell never gets a value and keeps `SkyLight 0`. `isLightOn` is written as `1` alongside it, which tells the client to trust the baked data instead of relighting. The propagation pass can't rescue it either, since `try_spread` skips any cell with opacity 15.

Those three are the only blocks Arnis places under open sky that are affected. I went through every name `block_definitions.rs` can emit against the current rules: `beetroots` and `torchflower_crop` happen to pass via the `"roots"` and `"flower"` substring matches, buttons and pressure plates are misclassified too but only ever sit on walls under a roof, and everything else is already covered.

Scope is narrow: `--bake-lighting` defaults to off and the GUI toggle is unchecked, so this only bites people who turned it on for Voxy or Chunky. Worth noting the bake happens against `DATA_VERSION 3955` (1.21.1), so opening the world in a newer version may relight it during the upgrade and hide the symptom.

The new test builds a farmland-and-wheat chunk, bakes it, and reads the nibble at the wheat cell. It fails with `left: 0, right: 15` when the three names are removed again.

## Oversized polygons render as a border around nothing

`flood_fill_area` measured `MAX_FLOOD_FILL_AREA` against the polygon's **bounding box** rather than the area it actually covers, and returned an empty vec past the cap. But callers paint the outline first and ask for the fill afterwards, e.g. `natural.rs` sets the edge blocks in its node loop and only then calls `get_or_compute`. So "no fill" didn't mean "not drawn", it meant an outline with untouched ground inside it.

A long diagonal coastline is the obvious shape that trips it: huge bounding box, small real area. 25M blocks is only 5000x5000, which is reachable on ordinary coastal and desert areas at 1:1.

Anything past the cap now goes through an even-odd scanline fill instead. It walks one Z row at a time, intersects it with each edge, sorts the crossings and records the span between each pair.

On the two constraints that matter here:

- **Memory.** No visited bitmap at all, which is the thing that made the bounding box the limit in the first place. Spans are counted in the same pass that finds them, so the output vector is allocated exactly once with no doubling. The cell budget stays at `MAX_FLOOD_FILL_AREA`, so worst-case output is the same as today's.
- **Time.** Cost is rows x edges, not area, and it's a single pass. The existing bitmap paths are untouched, so nothing below the cap changes and there's no cost on the normal path. The only polygons that do more work than before are the ones that currently produce nothing.

Two guards: `MAX_SCANLINE_EDGE_TESTS` refuses a pathological ring up front, and the per-span budget bails as soon as the count exceeds the cap, before allocating.

Semantics are even-odd on the outer ring, which matches what `geo::Contains` reports for the simple rings these are. Holes are already handled by callers subtracting a separate fill.

## Checks

`cargo fmt -- --check`, `cargo clippy --all-targets --all-features -- -D warnings` and `cargo test --all-targets --all-features` all pass. 429 tests, 5 new.

Both fixes were found while reading through the Teddy563/arnis changelog and then checking upstream. That fork carries its own versions of both, also under Apache-2.0.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/element_processing/leisure.rs
+++ b/src/element_processing/leisure.rs
@@ -5,7 +5,7 @@ use crate::deterministic_rng::element_rng;
 use crate::element_processing::bridges::BridgeSurfaceMap;
 use crate::element_processing::surfaces::get_blocks_for_surface;
 use crate::element_processing::tree::Tree;
-use crate::floodfill_cache::{BuildingFootprintBitmap, FloodFillCache};
+use crate::floodfill_cache::{is_oversized_ring, BuildingFootprintBitmap, FloodFillCache};
 use crate::osm_parser::{ProcessedMemberRole, ProcessedRelation, ProcessedWay};
 use crate::world_editor::WorldEditor;
 use rand::Rng;
@@ -49,6 +49,13 @@ pub fn generate_leisure(
             }
         }
 
+        // Resolve the fill before painting the edge, for the same reason as in natural.rs:
+        // a closed ring the fill refused must not leave a border around unfilled ground.
+        let filled_area = flood_fill_cache.get_or_compute(element, args.timeout.as_ref());
+        if filled_area.is_empty() && is_oversized_ring(element) {
+            return;
+        }
+
         // Process leisure area nodes
         for node in &element.nodes {
             if let Some(prev) = previous_node {
@@ -81,8 +88,6 @@ pub fn generate_leisure(
 
         // Flood-fill the interior of the leisure area using cache
         if corner_count > 0 {
-            let filled_area = flood_fill_cache.get_or_compute(element, args.timeout.as_ref());
-
             // Use deterministic RNG seeded by element ID for consistent results across region boundaries
             let mut rng = element_rng(element.id);
 
--- a/src/element_processing/natural.rs
+++ b/src/element_processing/natural.rs
@@ -4,7 +4,7 @@ use crate::bresenham::bresenham_line;
 use crate::deterministic_rng::element_rng;
 use crate::element_processing::bridges::BridgeSurfaceMap;
 use crate::element_processing::tree::{Tree, TreeType};
-use crate::floodfill_cache::{BuildingFootprintBitmap, FloodFillCache};
+use crate::floodfill_cache::{is_oversized_ring, BuildingFootprintBitmap, FloodFillCache};
 use crate::osm_parser::{ProcessedElement, ProcessedMemberRole, ProcessedRelation, ProcessedWay};
 use crate::world_editor::WorldEditor;
 use rand::{prelude::IndexedRandom, Rng};
@@ -145,6 +145,14 @@ pub fn generate_natural(
                 return;
             };
 
+            // Resolve the fill before painting the edge. It is cached, so this costs nothing
+            // extra. A closed ring that comes back empty is one the fill refused for size, and
+            // drawing its edge anyway leaves a border around ground nothing ever filled.
+            let filled_area = flood_fill_cache.get_or_compute(way, args.timeout.as_ref());
+            if filled_area.is_empty() && is_oversized_ring(way) {
+                return;
+            }
+
             // Process natural nodes to fill the area
             for node in &way.nodes {
                 let x: i32 = node.x;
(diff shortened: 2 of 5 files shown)


louis-e:
Pushed three follow-ups.

**Both review points taken.** `windows(2)` gives `len - 1` edges for a closed ring, so the work estimate was a row too pessimistic; and the span length now widens to `i64` before the subtraction instead of after. Neither was reachable, since `flood_fill_area` computes its own bbox area in `i32` first and would trip long before either, and `overflow-checks = true` would panic rather than silently corrupt the count. Both are free, so no reason not to.

**The caller side is now fixed too.** The scanline path covers the polygons that used to be dropped, but a ring can still be refused past the cell budget or the work cap, and `natural.rs` and `leisure.rs` paint the edge in their node loop before asking for the fill. So those cases still produced the border-around-nothing. Both now resolve the fill first (cached, so no extra work) and bail before painting.

The part worth being careful about: **empty on its own is not enough to bail on.** Three different things fill to nothing, and only one is a refusal.

- an open way, e.g. `natural=cliff` or `natural=ridge` traced as a line, which `flood_fill_area` rejects at the `first != last` guard. The bresenham line is the whole render there.
- a sliver too thin to contain a lattice point, so no seed lands inside.
- a ring the fill gave up on.

Only the third should lose its outline. `is_oversized_ring` requires a closed ring whose bbox is past `MAX_FLOOD_FILL_AREA`, which is the only path a size refusal can take, so the other two draw exactly as before. Four tests cover the distinction.

Worth flagging: Teddy563/arnis carries a version of this same fix that guards on `filled_area.is_empty() && way.nodes.len() > 2`. Their `flood_fill_area` has the same open-polyline rejection ours does, so as far as I can tell that guard drops every multi-node `natural=cliff` and `natural=ridge` entirely. Might be worth passing on.

`cargo fmt -- --check`, `cargo clippy --all-targets --all-features -- -D warnings` and `cargo test --all-targets --all-features` all pass. 433 tests, 9 new.

louis-e:
Taken, with one correction to the example.

The semantics point is right and now fixed. Sampling at `z + 0.5` and rounding spans outward meant this path decided occupancy differently from the bitmap paths it stands in for, and it leaned toward filling: a cell half covered by the polygon came out inside, and a span endpoint landing exactly on an edge was included. Rows are now sampled at integer `z` with the same half-open crossing rule, so a vertex on the row still counts once rather than twice, and spans exclude their endpoints so boundary cells are left to the caller's outline pass.

There is now a parity test that walks every lattice point of a concave ring through both this and `geo::Contains` and requires them to agree, 2744 points. The previous version failed it at `(1, 0)`, so the concern was real and the test bites.

**The wide-one-block-high triangle can't actually reach this path.** The scanline only runs when the bounding box is past 25M blocks, and a two-row-high shape would need to be over 12.5M blocks wide to qualify, which is 12500 km at 1:1. I added the case anyway as `a_one_block_high_wedge_never_overfills`, padded until it clears the cap, plus `a_zero_area_ring_fills_nothing_however_big_its_bbox` for a ring traced out and straight back. Both hold.

**One difference remains and I don't think it should be closed.** A cell lying exactly on a *horizontal* edge is invisible to ray casting, so this path can include it where `geo::Contains` would not. Excluding it needs a per-cell edge test, which is O(edges) per cell and exactly the cost this path exists to avoid. It is one block of slack on rings that are thousands of blocks across, the callers paint the boundary with bresenham anyway, and the alternative for these polygons is still the outline-around-nothing they render as today. Documented on the function rather than papered over.

On the framing generally: for polygons that reach this path the comparison is not "exact fill vs approximate fill", it is "approximate fill vs nothing at all". They are dropped entirely on main.

`cargo fmt -- --check`, `cargo clippy --all-targets --all-features -- -D warnings` and `cargo test --all-targets --all-features` all pass. 436 tests, 12 new.

PR #1283, opened 2026-08-18
Prepare v3.1.0

louis-e:


cukacukiellu-oss:
While testing the new Arnis v3.1.0 update, I noticed that the Signage setting doesn't save.

The update says that Arnis now remembers GUI settings, and most settings appear to persist correctly. However, the Signage option does not.

For example, if I select **Full** under Signage, close Arnis, and open it again, the setting resets to basic instead of remembering my previous selection.

This was found while testing Arnis v3.1.0.

Please fix it.

PR #1287, opened 2026-08-19
fix(barriers): use IRON_BARS for metal fences instead of STONE_BRICK_WALL

weed33834:
## What

The metal-fence family (`chain_link`, `metal`, `wire`, `barbed_wire`, `corrugated_metal`, `electric`, `metal_bars`) currently renders as `STONE_BRICK_WALL`, with a leftover `// IRON_BARS` comment marking the original intent (flagged by @matkoniecz on #1259).

This switches that branch to `IRON_BARS` (block id 36) and removes the stale comment.

## Why

`IRON_BARS` is the semantically correct block for metal fences and is already used across the codebase (`power.rs`, `amenities.rs`, `bridge_styles.rs`, …). Metal fences rendering as stone brick walls is a bug — iron bars are the right visual and semantic match.

## Scope

Single-line material change + comment removal. No other behavior change.

Follow-up to #1259 (`material=wood` → `OAK_FENCE`).

### DIFF
--- a/src/element_processing/barriers.rs
+++ b/src/element_processing/barriers.rs
@@ -39,7 +39,7 @@ pub fn generate_barriers(
                     "chain_link" | "metal" | "wire" | "barbed_wire" | "corrugated_metal"
                     | "electric" | "metal_bars",
                 ) => {
-                    barrier_material = STONE_BRICK_WALL; // IRON_BARS
+                    barrier_material = IRON_BARS;
                     barrier_height = 2;
                 }
                 Some("slatted" | "paling") => {


weed33834:
Hi maintainers — heads up on the red **benchmark** check: the benchmark itself passed cleanly (generation 21s vs baseline 18s = unchanged; peak memory 1149 MB vs baseline 1160 MB = unchanged; no regression from a one-line material swap).

The failure is infrastructure-only: the workflow's `thollander/actions-comment-pull-request` step uses the default `GITHUB_TOKEN`, which is read-only for pull requests from forks, so posting the benchmark report comment is rejected (`Resource not accessible by integration`). `build` and `lint` are both green.

Happy to rebase or adjust if anything else is needed.

PR #1297, opened 2026-08-25
Update landuse.rs

BOBJOEx11g:
added support for salt flats and salt ponds, as diorite


BOBJOEx11g:
checks failed because . instead of , after diorite, needs to be fixed

BOBJOEx11g:
I fixed any errors with my code. All done.

PR #1300, opened 2026-08-28
Add concrete powder colors to map preview palette

benjaminpjones:
### Problem

`DEFAULT_ROAD_MIX` in `src/element_processing/highways.rs` is:

```rust
const DEFAULT_ROAD_MIX: &[Block] = &[GRAY_CONCRETE_POWDER, CYAN_TERRACOTTA];
```

so `gray_concrete_powder` is the primary road surface for most highway types. It has no entry in `get_block_colors()`, so it falls through this branch of `get_fallback_color()`:

```rust
if name.contains("concrete") {
    return Rgb([128, 128, 128]);
}
```

which returns exactly the same value as `("stone", Rgb([128, 128, 128]))`.

In `--map-preview` output, asphalt is therefore byte-identical to stone. The only thing separating a road from a stone building afterwards is `apply_elevation_shading`'s multiplier, and its range overlaps between the two — so the two are not reliably separable by color at all.

### Fix

Add the full 16-block concrete powder family to `get_block_colors()`, next to the existing concrete entries.

Values are mean texture colors — the same convention as the surrounding entries. I verified the convention by recomputing the committed `gray_concrete`, `white_concrete`, `black_concrete` and `red_concrete` values from their textures; all four reproduce exactly (delta 0 on every channel), so the new values are derived the same way rather than eyeballed.

`cargo test --release --no-default-features` — 611 passed, 0 failed.

Only affects preview rendering; no change to world generation or block placement.

### Manual testing

Place Charles de Gaulle, Paris

```bash
cargo build --release --no-default-features
./target/release/arnis \
  --bbox=48.868051,2.286259,48.879549,2.303741 \
  --scale 0.3 --mode geo-only --map-preview \
  --output-dir /tmp/etoile
```

| before (`main`) | after (this PR) | diff
|---|---|---|
|  |  |  |


louis-e:
Thanks, LGTM!

PR #1309, opened 2026-09-01
Add Georgian localization

TGaming03:
This PR adds Georgian (ka-GE) localization to Arnis.

I also added ka-GE to the locale test list.


louis-e:
Thanks, looks good to me! :)

PR #1316, opened 2026-09-02
Fix invalid biome data

### LINKED ISSUE #1296: [BUG] Invalid biome data
**Describe the bug**
A user reported that they were not able to open a world created by this program in Amulet.
I believe you are writing the biome data incorrectly but the game seems to be handling it.

Here is some good documentation on the format
https://team-lodestone.github.io/Documentation/Bedrock/LevelDB_Output_Array_Formats/Data3D_Format.html

Here is where your code writes the biome data.
https://github.com/louis-e/arnis/blob/c7b5f19de05488bd401bb32fc4c0eb3baf151e23/src/world_editor/bedrock.rs#L623

I believe it should either be a multiple of `\xFF` (null section) or `\x00\x00\x00\x00\x00`.
Alternatively you can omit the whole Data3D entry and the game seems to populate it correctly when it is first loaded.

gentlegiantJGC:
A biome section with value 0 is followed by a 32 bit integer containing the biome value for that sub-chunk.
This means the biome data must be a multiple of 5 bytes.
The data as this program writes it is an ocean biome for the first 5 sub-chunks and only has half of the 6th sections 32 bit integer. 
The game seems to gracefully handle this but other tools do not.
255 or \xFF is the value meaning there is no data for that section.

Fixes #1296


gentlegiantJGC:
Looks like the tests passed but your CI broke

louis-e:
retrigger-benchmark

louis-e:
Thanks for the contribution!

PR #1340, opened 2026-09-09
Update pr-benchmark.yml

Emmett105:


Emmett105:
if this is approved it will add a installation guide for a few linux distros including fedora arch and ultramarine and it includes a guide for macos and linux

